Issue #5, opened 2020-12-20
Implement ICurrentAllEventsQuery


to11mtm:
We don't currently have `ICurrentAllEventsQuery` Implemented for Linq2Db Journal. We should add this functionality so that users who require this functionality can migrate.

to11mtm:
Closed by #10

PR #10, opened 2020-12-26
Add IAllEventsQuery, ICurrentAllEventsQuery

### LINKED ISSUE #4: Implement IAllEventsQuery
We don't currently have `IAllEventsQuery` implemented on Linq2Db journal. Some Persistence.Query users may require this functionality to migrate over.

to11mtm:
Closes #4 and #5 

- Added Support for IAllEventsQuery
- Added Support for ICurrentAllEventsQuery
- Cleaned up attribute mapping for JournalRow, moved all to code mapping for consistency
- Cleaned up a few minor comments/etc.
- Cleaned up exposed Persistence.Query `Source`s such that the Query should not be run until the stream is run.
- Cleaned up Persistence.Query specs; they were passing however the use of table names was incorrect.


src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -390,32 +392,34 @@ public override
                 }
 
                 
-                //TODO: Is there a better way to do this async?
-                //return Source
-                //    .FromObservable(query.AsAsyncEnumerable().ToObservable())
-                
-                return Source.From(await query.ToListAsync())
+                return AsyncSource<JournalRow>.FromEnumerable(query,async q=>await q.ToListAsync())
```
to11mtm, line 395: The idea with this change is that Streams should be 'lazy'. While in most cases this specific method is safely called from inside another stream's stage, keeping things consistent seemed worth it.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -390,32 +392,34 @@ public override
...
-                            return new
-                                Util.Try<ReplayCompletion>(
-                                    sertry.Failure.Value);
-                        }
-                    });
+                        deserializeFlow).Select(MessageWithBatchMapper());
             }
         }
+
+        private static Func<Util.Try<(IPersistentRepresentation, IImmutableSet<string>, long)>, Util.Try<ReplayCompletion>> MessageWithBatchMapper()
```
to11mtm, line 401: Somehow, this was chosen by refactoring tools as the 'best' way to refactor with minimal allocations. I'm not sure that's true.

src/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalRow.cs
```diff
@@ -5,20 +5,15 @@ namespace Akka.Persistence.Sql.Linq2Db.Journal.Types
...
 
         public bool deleted { get; set; }
-        //[Column(Configuration = ProviderName.SQLite, IsPrimaryKey = false)]
-        //[Column(IsPrimaryKey = true, CanBeNull = false)]
         public string persistenceId { get; set; }
-        //[Column(Configuration = ProviderName.SQLite, IsPrimaryKey = false)]
-        //[Column(IsPrimaryKey = true)]
+        
         public long sequenceNumber { get; set; }
-        [Column(CanBeNull = false)]
```
to11mtm, line 21: Moved this to the code mapping

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -42,17 +48,33 @@ protected IQueryable<JournalRow> baseQuery(DataConnection connection)
                 .Where(jr =>
                     includeDeleted == false || (jr.deleted == false));
         }
+        
+        protected static IQueryable<JournalRow> baseQueryStatic(DataConnection connection, bool includeDeleted)
```
to11mtm, line 52: Needed a static version of this to avoid captures

src/Akka.Persistence.Sql.Linq2Db/Query/Linq2DbReadJournal.cs
```diff
@@ -273,12 +299,98 @@ IEnumerable<string> next(string id)
...
+            {
+                return new[]{ await input.readJournalDao.MaxJournalSequenceAsync()};
+            }).ConcatMany(
+                maxInDb =>
+                {
+                    return _eventsByTag(tag, offset, Some.Create(maxInDb));
+                });
+        }   
+        
+        private Source<EventEnvelope, NotUsed> _events(
```
to11mtm, line 313: There is probably some opportunity here for re-use. `_events` and `_eventsByTag` are nearly identical.

PR #12, opened 2021-01-18
migrated build scripts to install .NET Core 3.1

Aaronontheweb:
This should help resolve some of the issues the CI system was having when I first ran it


Aaronontheweb:
CI can run now @to11mtm but it looks like there's a number of problems with the test suite - the usual stuff: runtime vs. DLL conflicts.

PR #22, opened 2021-05-02
Serialization pipeline Fixes/improvements

to11mtm:
- Fix an Edge case around recovering Old Journal entries that did not include a Serializer Identifier.
- Improved `NumericRangeEntry` behavior to use a `yield` for it's Enumerator instead of allocating a bunch of lists.
- Use new `TState` overloads For Seriallizer `.WithTransport` methods to improve copies just a little
- Make sure we are always transacting for batched writes so we correctly on all edge cases


to11mtm:
@Aaronontheweb Looks like Docker tests failing somehow =/. Confused cause I'm not sure what I broke?

PR #29, opened 2021-08-18
Fix issue where Persist failures were being reported as rejects.

to11mtm:
Also Use an enum instead of a class on flowcontrol.

### DIFF
--- a/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Data;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
@@ -138,21 +139,19 @@ namespace Akka.Persistence.Sql.Linq2Db.Journal.DAO
             {
                 //hot path:
                 //If we only have one row, penalty for BulkCopy
-                //Isn't worth it due to insert caching/etc.
-                if (xs.Count > 1)
-                {
+                //Isn't worth it due to insert caching/transaction/etc.
+                var count = xs.Count;
+                if (count is > 1)
                     await InsertMultiple(xs);
-                }
-                else if (xs.Count > 0)
-                {
-                    await InsertSingle(xs);
-                }
+                else if (count == 1) await InsertSingle(xs);
             }
 
         }
 
         private async Task InsertSingle(Seq<JournalRow> xs)
         {
+            //If we are writing a single row,
+            //we don't need to worry about transactions.
             using (var db = _connectionFactory.GetConnection())
             {
                 await db.InsertAsync(xs.Head);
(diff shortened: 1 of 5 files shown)


src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -200,45 +195,61 @@ await db.GetTable<JournalRow>()
...
-                TaskScheduler.Default);
+            //Fold our List of Lists into a single sequence
+            var rows = Seq(FlattenListOfListsToList(serializedTries));
+            //Wait for the write to go through. If Task fails, write will be captured
+            //As WriteMessagesFailure.
+            await QueueWriteJournalRows(rows);
+            //If we get here, we build an ImmutableList containing our rejections.
+            //These will be captured as WriteMessagesRejected
+            return BaseByteArrayJournalDao
+                .BuildWriteRejections(serializedTries);
```
to11mtm, line 235: We were doing the wrong thing before, Plz doublecheck against [original scala](https://github.com/akka/akka-persistence-jdbc/blob/63238b87f26b60af5c72494211a5f8943ef6ff66/core/src/main/scala/akka/persistence/jdbc/journal/dao/legacy/ByteArrayJournalDao.scala#L73-L86).

The idea here -should- be that if QueueWriteJournalRows fails, or the write fails, we throw here and generate Failure. Otherwise, we will see if any of the writes failed serialization and treat those as Rejected.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -138,21 +139,18 @@ private async Task WriteJournalRows(Seq<JournalRow> xs)
...
+                //Isn't worth it due to insert caching/transaction/etc.
+                var count = xs.Count;
+                if (count > 1)
                     await InsertMultiple(xs);
-                }
-                else if (xs.Count > 0)
-                {
-                    await InsertSingle(xs);
-                }
+                else if (count == 1) await InsertSingle(xs);
```
Aaronontheweb, line 146: LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: Makes sense to me - the rows that go through should be a success but the ones that encountered failures should get a rejected message

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -370,18 +394,19 @@ await db.GetTable<JournalMetaData>()
...
-                        {
-                            SequenceNumber = md.SequenceNumber,
-                            PersistenceId = md.PersistenceId
-                        }));
+                queryable = queryable.Union(nextQuery.Select(metaDataSelector));
             }
 
-            return queryable.OrderByDescending(r => r.SequenceNumber)
-                .Select(r => r.SequenceNumber).Take(1);
+            return queryable.OrderByDescending(sequenceNumberSelector)
```
Aaronontheweb, line 400: LGTM on allocations

Aaronontheweb (approved):
LGTM

PR #47, opened 2021-12-14
Changeup behavior of MaxSequenceNumber to use Max wherever possible.

to11mtm:


src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -374,31 +374,75 @@ await db.GetTable<JournalMetaData>()
...
-            
-            
-
-            var queryable = db.GetTable<JournalRow>()
-                .Where(r => r.persistenceId == persistenceId).Select(rowDataSelector);
             if (minSequenceNumber != 0)
             {
-                queryable = queryable.Where(r =>
-                    r.SequenceNumber > minSequenceNumber);
+                if ((_journalConfig.DaoConfig.SqlCommonCompatibilityMode))
```
to11mtm, line 382: So, this was kinda ugly to make it work everywhere in a way that:

 - Gave reasonably Clean/Sane SQL in all scenarios
 - Had C# code that was reasonably sane to read
 - Wasn't disgusting in IL

Based on testing, having separate methods like such gives us more consistent performance in recovery, and is not that much worse to maintain.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -374,31 +374,75 @@ await db.GetTable<JournalMetaData>()
...
+                return MaxSeqForPersistenceIdQueryableNativeMode(db, persistenceId);
+            }
+        }
+
+        private static IQueryable<long?> MaxSeqForPersistenceIdQueryableNativeMode(DataConnection db,
+            string persistenceId)
+        {
+            return db.GetTable<JournalRow>()
+                .Where(r => r.persistenceId == persistenceId)
+                .Select(r => (long?)r.sequenceNumber);
```
to11mtm, line 407: So, to make best SQL we have to do a `.MaxAsync()` in the main call (`.FirstOrDefault` results in `TAKE` semantics and additional param pass), so for this case we don't do any constraining.

This is the happy path where users are fully on linq2db and not running compatibility mode. We will wind up with SQL like:

```sqlite
SELECT
	Max([r].[sequence_number])
FROM
	[journal] [r]
WHERE
	[r].[persistence_id] = @persistenceId
```

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -374,31 +374,75 @@ await db.GetTable<JournalMetaData>()
...
+        }
 
-            return queryable.OrderByDescending(sequenceNumberSelector)
-                .Select(sequenceNumberSelector).Take(1);
+        private static IQueryable<long?> MaxSeqForPersistenceIdQueryableCompatibilityMode(
+            DataConnection db, string persistenceId)
+        {
+            return db.GetTable<JournalRow>()
+                .Where(r => r.persistenceId == persistenceId).Select(r =>
+                    LinqToDB.Sql.Ext.Max<long?>(r.sequenceNumber).ToValue()).Union(db
```
to11mtm, line 441: So, there is 'technically' a way to call this version of `.Max()` via Extension method, but of course it collides with `IEnumerable<T>.Max()`. We are coercing the nullable long here so that we can `.GetValueOrDefault()` in the main call.

This is 'Compatibility mode' where Deletes are compatible with Sql.Common. We wind up getting something like below. There -might- be a better way to write the query and avoid the parameter dupe.

```sqlite
SELECT
	Max([t3].[c1])
FROM
	(
		SELECT
			[t1].[c1]
		FROM
			(
				SELECT
					MAX([r].[sequence_number]) as [c1]
				FROM
					[journal] [r]
				WHERE
					[r].[persistence_id] = @persistenceId
			) [t1]
		UNION
		SELECT
			[t2].[c1]
		FROM
			(
				SELECT
					MAX([r_1].[sequenceNr]) as [c1]
				FROM
					[journal_metadata] [r_1]
				WHERE
					[r_1].[PersistenceId] = @persistenceId_1
			) [t2]
	) [t3]
```

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -436,9 +480,9 @@ await db.GetTable<JournalRow>()
         {
             using (var db = _connectionFactory.GetConnection())
             {
-                return await MaxSeqNumberForPersistenceIdQuery(db,
+                return (await MaxSeqNumberForPersistenceIdQuery(db,
                     persistenceId,
-                    fromSequenceNr).FirstOrDefaultAsync();
+                    fromSequenceNr).MaxAsync()).GetValueOrDefault(0);
```
to11mtm, line 485: Linq2Db Expects that `.MaxAsync()` on `long` or `int` types will return a value; otherwise an exception is thrown. So we do this instead on the `IQueryable<long?>`.

src/Akka.Persistence.Sql.Linq2Db.Tests/Sqlite/MSSQLiteJournalSpec.cs
```diff
@@ -7,6 +7,12 @@
 
 namespace Akka.Persistence.Sql.Linq2Db.Tests
 {
+    public class MSSQLiteNativeConfigSpec : MSSQLiteJournalSpec
```
to11mtm, line 10: I noticed that we didn't check all SQL against native config in *actual* test. Some of the benches, however, covered the scenario. So, this never let breaks slip through (as issues would be caught in perf part of CI), but now we have a more clear suite for ourselves and contributors to make sure everything works the way they expect.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: I think it's probably fine

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: Looks just like what we have in Akka.Persistence.Sql.Common - I think it's fine. Thanks for spelling out the generated SQL.

Aaronontheweb (approved):
LGTM

PR #68, opened 2022-03-19
Tag table with backwards compatibility

### LINKED ISSUE #67: Feature: Store Tags in separate table rather than inline
**Is your feature request related to a problem? Please describe.**

Users would like the ability to use a separate table for tags with SQL Persistence, as described in https://github.com/akkadotnet/akka.net/issues/5296 . 

Additionally, Existing users of Akka.Persistence.Linq2Db would like to ensure they continue to maintain fast read and write performance from their existing journals.

It would be extra great, if we could provide users a way to migrate with zero downtime (if more effort.)

**Describe the solution you'd like**

We can leverage Linq2Db and Akka.Persistence.Linq2Db's pipeline to provide a solution. 

By providing 3 Read and Write modes, we can allow for a 'seamless' deployment, ideal state would be:

 - Writers and readers are in 'old' mode.
 - Deploy code for writers and readers in 'migration' mode. Writers will write both the comma separated tags and to the tag table, Readers will check for tags from both.
 - Run utility or DB Script to put old tags into the tag table.
 - Deploy code for writers and readers in 'new' mode.
 - DBAs can clear out tag column if needed.

to11mtm:
Fixes #67 

(Alternative solution to https://github.com/akkadotnet/akka.net/issues/5296)

## Changes

This is an implementation of Tag tables with a form of backwards compatibility.

- Added `TagWriteMode` enum, with 3 modes of operation.
  - applied logic to `ByteArrayJournalSerializer` so that tags may be applied based on setting
  - Added logic to `BaseByteArrayJournalDao` to switch behavior based on mode of operation
- Added logic to `BaseByteArrayJournalDao` to interleave batch-untagged and per-row-tagged writes when writing to tag table

- Added `TagReadMode` enum, with 3 modes of operation.
  - Added logic to `BaseByteReadArrayJournalDao` to respect enumeration
    - Queries running off tags will check tag table and/or main table based on setting
    - Filling of tags for stream pipeline will respect setting

## Todo:

 * [ ] Some cleanup in the Read Journal and elsewhere
 * [ ] Config bits (defaults, reference, table/column names)
 * [ ] Create table bits
 * [ ] Tests around new functionality
   - Ideally, we should have tests around migration scenarios. 
 * [ ] Update deletion logic to delete tags in tag table
   - *Note:* a combination of 'delete-compatibility-mode' and tag-tables may make deletions very painful, since you'd be locking across 3 tables.
 * [ ] Update 'update' functionality so that it will update tag table if needed.
   - the update functionality is a non-standard addition present from persistence-jdbc.

## Not in scope of this PR:

 - Migration routines/scripts. I think a helper application similar to what has been proposed for indexes would be a good idea, but this is likely a big enough PR as-is.

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks


src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -148,6 +154,52 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
...
+            if (config.TableConfig.UseEventManifestColumn)
+            {
+                journalRowBuilder.Member(r => r.eventManifest)
+                    .IsColumn().HasLength(64);
+            }
+            else
+            {
+                journalRowBuilder.Member(r => r.eventManifest)
+                    .IsNotColumn();   
+            }
```
to11mtm: Basically, this switch should control whether the 'fixed' behavior for eventManifest is used.

If it's not mapped, the property will come back null and the other logic works as it should, and is ignored on inserts.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -23,6 +24,37 @@
 
 namespace Akka.Persistence.Sql.Linq2Db.Journal.DAO
 {
+    public class SequentialUUIDGenerator
```
to11mtm: This algo is borrowed from EF Core, and should be attributed such and re-commented for our use.

The expectation is that we should be able to generate Guid-esque structures with some 'predictability' in a range of output.

This function's use may require fine tuning to produce optimal results.

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -100,53 +103,271 @@ private static int MaxTake(long max)
...
+                                    r.ordering <= input.maxOffset)
+                                .Take(input.maxTake).ToListAsync();
+                            await addTagDataFromTagTable(mainRows, conn);
+                            return mainRows;
+                        }
+                    }).Via(perfectlyMatchTag(tag, separator))
+                .Via(deserializeFlow);
+        }
+
+        private Expression<Func<JournalRow, bool>> eventsByTagMigrationPredicate(DataConnection conn, string tagVal)
```
to11mtm: This had to be expanded into it's own method because of the compiler version used in build.

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -100,53 +103,271 @@ private static int MaxTake(long max)
...
+                            return mainRows;
+                        }
+                    })
+                //We still PerfectlyMatchTag here
+                //Because DB Collation :)
+                .Via(perfectlyMatchTag(tag, separator))
                 .Via(deserializeFlow);
+        }
 
+        private Expression<Func<JournalRow, JournalTagRow, bool>> EventsByTagOnlyJoinPredicate
```
to11mtm, line 219: This had to be expanded as such due to build server compiler not being able to infer expressions being returned from a ternary.

Aaronontheweb:
Going to review this, as I am very interested in having this functionality for a project.

src/Akka.Persistence.Sql.Linq2Db/Config/BaseByteArrayJournalDaoConfig.cs
```diff
@@ -43,6 +44,6 @@ public BaseByteArrayJournalDaoConfig(Configuration.Config config)
         public int BufferSize { get; protected set; }
         
         public bool SqlCommonCompatibilityMode { get; protected set; }
-        
+        public int DbRoundTripTagBatchSize { get; set; }
```
Aaronontheweb: How does this get used exactly?

src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableConfig.cs
```diff
@@ -3,6 +3,19 @@
 
 namespace Akka.Persistence.Sql.Linq2Db.Config
 {
+    [Flags]
+    public enum TagWriteMode
```
Aaronontheweb, line 6: Got it - this is how we can support data written in previous journal implementations.

src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableConfig.cs
```diff
@@ -13,6 +26,12 @@ public class JournalTableConfig
         public string MetadataTableName { get; protected set; }
         public MetadataTableColumnNames MetadataColumnNames { get; protected set; }
         public bool WarnOnAutoInitializeFail { get; }
+        
+        public TagWriteMode TagWriteMode { get; }
+        public TagTableMode TagTableMode { get; }
+        public string? TagTableName { get; }
+        public bool UseEventManifestColumn { get; }
```
Aaronontheweb: These are all of the aforementioned compat-flags?

src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableConfig.cs
```diff
@@ -28,6 +47,31 @@ public JournalTableConfig(Configuration.Config config)
...
+                res = TagWriteMode.CommaSeparatedArrayAndTagTable;
+            }
+            else if (s.Equals("tagtableonly",
+                         StringComparison.InvariantCultureIgnoreCase))
+            {
+                res = TagWriteMode.TagTable;
+            }
+            else
+            {
+                res = TagWriteMode.CommaSeparatedArray;
```
Aaronontheweb: Got it, this is probably how most popular platforms have handled this.

src/Akka.Persistence.Sql.Linq2Db/Config/ReadJournalPluginConfig.cs
```diff
@@ -7,11 +9,35 @@ public ReadJournalPluginConfig(Configuration.Config config)
...
 
         public string Dao { get; set; }
 
         public string TagSeparator { get; set; }
+        public TagReadMode TagReadMode { get; set; }
+        public TagTableMode TagTableMode { get; }
+    }
+
+    [Flags]
+    public enum TagReadMode
```
Aaronontheweb, line 28: How often will a journal have a separate read / write mode? Only during a migration?

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -135,7 +135,13 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
                 .HasColumnName(tableConfig.ColumnNames.SequenceNumber)
                 .Member(r => r.Timestamp)
                 .HasColumnName(tableConfig.ColumnNames.Created);
-
+            //We can skip writing tags the old way by ignoring the column in mapping.
+            journalRowBuilder.Member(r => r.tagArr).IsNotColumn();
```
Aaronontheweb: LGTM

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -148,6 +154,52 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
                     .Member(r=>r.sequenceNumber).IsPrimaryKey();
             }
 
+            void SetJoinCol<T>(PropertyMappingBuilder<JournalTagRow, T> builder,
+                PropertyMappingBuilder<JournalRow, long> propertyMappingBuilder)
+            {
+                if (config.TableConfig.TagTableMode ==
+                    TagTableMode.SequentialUUID)
+                {
+                    builder.Member(r => r.JournalOrderingId)
```
Aaronontheweb: LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: Does https://github.com/akkadotnet/akka.net/pull/5995 help with this?

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -149,34 +184,145 @@ private async Task WriteJournalRows(Seq<JournalRow> xs)
...
+        private async Task InsertWithOrderingAndBulkInsertTags(DataConnection dc,
+            Seq<JournalRow> xs)
+        {
+            var tagsToInsert = new List<JournalTagRow>(xs.Count);
+            foreach (var journalRow in xs)
+            {
+                var dbid = await dc.InsertWithInt64IdentityAsync(journalRow);
+                foreach (var s1 in journalRow.tagArr)
+                {
+                    tagsToInsert.Add(new JournalTagRow(){JournalOrderingId = dbid, TagValue = s1});
```
Aaronontheweb: LGTM

Aaronontheweb (commented):
Have a partial review for you @to11mtm - have about half the files to look at still.

ismaelhamed:
Is this a different approach than [#467](https://github.com/akka/akka-persistence-jdbc/pull/467)?

to11mtm:
> Is this a different approach than [#467](https://github.com/akka/akka-persistence-jdbc/pull/467)?

Yes and not quite:

Yes:

 - If you are running based off Ordering, it is close to the same approach as 467, with the added kindness of having migration modes. (The migration modes add a good bit of complexity here and are where much of the deviation occurs.)

Not Quite:

 - We aren't subclassing the way JDBC does to handle the different behavior modes. Given we are trying to provide migration modes, it would have meant 3 classes and a more complex migration process.
 - We are adding the option of using some form of SequentialUUID for joining to the tag table rather than orderingID.
   - This is a mode intended for scenarios where high write throughput with tags is desired, at the expense of additional storage and possibly (likely) slower reads.
   - This may not be a good idea, or may require additional optimization
 - We may or may not be doing the Inserts the same way SlickDB does. Based on the initial PRs for Tag tables in persistence-jdbc, I'm -guessing- their approach was similar to ours (where you have to insert each row one at a time to get IDs). Assumed this is still true based on them having split logic similar to this OR, where if there are no tags they try to do a simpler write.
   - To solve this, Linq2Db will need to have multirow insert-with-return support, Or https://github.com/linq2db/linq2db/issues/2960 needs to be resolved.

src/Akka.Persistence.Sql.Linq2Db/Config/BaseByteArrayJournalDaoConfig.cs
to11mtm: I thought I left comments, IDK where then went, so just in case...

 - This controls the max number of Tags we will round-trip in a batch.
   - Tags are small compared to message payloads in most cases, so it's OK to do more tags vs Rows in a round trip I would think.

src/Akka.Persistence.Sql.Linq2Db/Config/ReadJournalPluginConfig.cs
to11mtm: > Only during a migration?

I would hope so. The 'both' modes are (probably) not at all performant and you'd only want to use if you were in a 'We must have zero downtime' situation and wanted to migrate.

The only other scenario I could think of is if you have some other 'things' that are doing special stuff against your existing tag column. (i.e. querying against the old tag column in some form of reporting/diagnostic query/function/sproc.) In that case they'd have to migrate that 'thing' as well.

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
to11mtm: probably!

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
to11mtm: @Aaronontheweb TBH we could really use more tests around things, but I have never used tags so I have a hard time coming up with scenarios.

Aaronontheweb:
@to11mtm I'll review this again this week. Had it on my daily to-do list for a while but would really like to see it merged and Akka.Persistence.Linq2Db become the new standard ;)

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
Aaronontheweb: LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: Can we use https://github.com/akkadotnet/akka.net/pull/5995 for this instead? I'd prefer to keep 1 implementation floating around, but if that approach doesn't make sense then no worries.

Aaronontheweb (approved):
LGTM - however, for testing this prior to going into production I'd like to propose the following:

1. Create 3 databases: Akka.Persistence.SqlServer, Postgres, and MySQL with pre-populated data using tags - all created from the same application that uses the same persistent queries. Save those DB files and commit them into a repo somewhere.
2. Build a docker image that mounts and restores those databases.
3. Run an application that uses Akka.Persistence.Linq2Db's implementation of the journal and see if we can re-run the existing queries and produce the same results against all 3.
4. Re-run a version of this using custom object serialization, just to ensure that is also supported in the migration.

That would pass the snuff test for me to a degree where I'd feel comfortable deprecating Akka.Persistence.Sql.Common and its subsequent implementations. Our team at Petabridge can help with this. What do you think @to11mtm @ismaelhamed ?

Aaronontheweb:
Build is failing due to NuGet.org errors at the moment.

Aaronontheweb:
@to11mtm what do you think of my review comment above? Should we mark this as "ready" and merge it in, with a plan to complete those above steps with help from our team?

to11mtm:
> LGTM - however, for testing this prior to going into production I'd like to propose the following:

It's not done. That TODO list at the top for starters.

> 
>     1. Create 3 databases: Akka.Persistence.SqlServer, Postgres, and MySQL with pre-populated data using tags - all created from the same application that uses the same persistent queries. Save those DB files and commit them into a repo somewhere.
> 

You can't do MySQL until https://github.com/akkadotnet/Akka.Persistence.Linq2Db/issues/3 is completed.

>     2. Build a docker image that mounts and restores those databases.
> 
>     3. Run an application that uses Akka.Persistence.Linq2Db's implementation of the journal and see if we can re-run the existing queries and produce the same results against all 3.
> 
>     4. Re-run a version of this using custom object serialization, just to ensure that is also supported in the migration.
> 
> 
> That would pass the snuff test for me to a degree where I'd feel comfortable deprecating Akka.Persistence.Sql.Common and its subsequent implementations.

The other filed issues should be considered before deprecating Sql.Common, timestamp generator is one that comes to mind.

> Our team at Petabridge can help with this. What do you think @to11mtm @ismaelhamed ?

As I've mentioned in other conversations this is not a task I've had the bandwidth to take over the finish line, both in scope of the TODOs (bandwidth) as well as more full testing (I'm not familiar enough with query to come up with good tests). I can answer questions and advise but as it stands the bus factor on this is too low.

As another side note, originally we hoped to deprecate sql.common earlier. As it stands, we risk a large influx of support issues and will risk more noise between compatibility issues of switching from sql.common vs whatever set of compatibility flags they choose (i.e. eventmanifest but no tag table, compatibility modes on different DBs... delete surprises...) something to consider.

tl;dr- TODOs need to be done before this gets merged, until then the tests you describe give a false sense this is done.

Aaronontheweb:
@Arkatufus you should move your commits onto its own PR and try to leave @to11mtm 's work here intact, for the sake of making it easy to compare the difference between the two separately.

src/Akka.Linq2Db.Sandbox/Akka.Linq2Db.Sandbox.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+    <PropertyGroup>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Akka.Streams" Version="1.4.21" />
+      <PackageReference Include="linq2db" Version="3.4.1" />
+      <PackageReference Include="Reactive.Streams" Version="1.0.2" />
```
Aaronontheweb, line 11: No need for an explicit Reactive.Streams reference here

src/Akka.Persistence.Linq2Db.Compatibility.Tests/Akka.Persistence.Linq2Db.Compatibility.Tests.csproj
```diff
@@ -9,6 +9,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="Akka.TestKit" Version="$(AkkaVersion)" />
         <PackageReference Include="Akka" Version="$(AkkaVersion)" />
```
Aaronontheweb, line 13: Can remove explicit `Akka` reference if we're referencing the TestKit

src/Akka.Persistence.Sql.Linq2Db/Config/BaseByteArrayJournalDaoConfig.cs
```diff
@@ -9,6 +9,7 @@ public BaseByteArrayJournalDaoConfig(Configuration.Config config)
             BufferSize = config.GetInt("buffer-size", 5000);
             BatchSize = config.GetInt("batch-size", 100);
             DbRoundTripBatchSize = config.GetInt("db-round-trip-max-batch-size", 1000);
+            DbRoundTripTagBatchSize = config.GetInt("db-round-trip-max-tag-batch-size", 1000);
```
Aaronontheweb, line 12: LGTM

src/Akka.Persistence.Sql.Linq2Db/Config/TagTableConfig.cs
```diff
@@ -0,0 +1,41 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="TagTableConfig.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
```
Aaronontheweb, line 3: Can get rid of Lightbend attribution here.

src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
```diff
@@ -0,0 +1,43 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="TagTableColumnNames.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
```
Aaronontheweb, line 3: Same here

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -140,6 +140,13 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
                 .Member(r => r.Timestamp)
                 .HasColumnName(columnNames.Created);
 
+            //We can skip writing tags the old way by ignoring the column in mapping.
+            journalRowBuilder.Member(r => r.TagArr).IsNotColumn();
+            if (tableConfig.TagWriteMode == TagWriteMode.TagTable)
```
Aaronontheweb, line 145: Based on @to11mtm 's suggestion, we're going to need to be able to support multiple tag write modes. Can we address this in a subsequent PR?

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -15,12 +15,44 @@
...
 using Akka.Streams.Dsl;
 using LanguageExt;
-using LinqToDB;
+using LinqToDB;using LinqToDB.Common;
 using LinqToDB.Data;
 using static LanguageExt.Prelude;
 
 namespace Akka.Persistence.Sql.Linq2Db.Journal.Dao
 {
+    public class SequentialUuidGenerator
```
Aaronontheweb, line 24: Can make this `static`

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/ByteArrayJournalSerializer.cs
```diff
@@ -18,13 +19,15 @@ public sealed class ByteArrayJournalSerializer : FlowPersistentReprSerializer<Jo
         private readonly string _separator;
         private readonly IProviderConfig<JournalTableConfig> _journalConfig;
         private readonly string[] _separatorArray;
+        private readonly TagWriteMode _tagWriteMode;
```
Aaronontheweb, line 22: Consideration for multiple write modes here as well

src/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalRow.cs
```diff
@@ -1,8 +1,16 @@
...
 
 namespace Akka.Persistence.Sql.Linq2Db.Journal.Types
 {
+    public sealed class JournalTagRow
+    {
+        public long JournalOrderingId { get; set; }
+        
+        public string TagValue { get; set; }
+        
+        public Guid WriteUuid { get; set; }
```
Aaronontheweb, line 11: No need to include `WriteUuid` on the column - OrderingId should be good enough if it's unique.

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -129,6 +131,15 @@
             sequence-number = sequence_number
           }
         }
+        
+        tag {
+          table-name = "tags"
+          columns {
+            ordering-id = ordering_id
+            tag-value = tag
+            writer-uuid = writer_uuid
```
Aaronontheweb, line 140: Can remove this

Aaronontheweb (approved):
Approving and merging this because it's a blocker and I don't want to review another 90+ file PR again, but there are some issues that need to be addressed in the comments. Let's deal with those in separate PRs.

PR #107, opened 2022-10-31
Update common.props usage and upgrade all package versions

Arkatufus:
## Changes

* Update common.props file
* Update .csproj files
* Bump all versions in common.props

### DIFF
--- a/Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests/Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests.csproj
+++ b/Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests/Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests.csproj
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="..\src\common.props" />
 
     <PropertyGroup>
         <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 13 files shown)


Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests/Akka.Persistence.Linq2Db.Benchmark.DockerComparisonTests.csproj
```diff
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="..\src\common.props" />
```
Aaronontheweb, line 2: We can skip explicitly referencing `common.props` if we just name the file `Directory.Build.props`, which is probably how we should do this going forward.... BUUUT dependabot may not like it.... So we'll stick with this for now.

build-system/pr-validation.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     parameters:
       name: 'windows_pr'
       displayName: 'Windows PR Validation'
-      vmImage: 'vs2017-win2016'
+      vmImage: 'windows-2019'
```
Aaronontheweb, line 21: `windows-latest`

Aaronontheweb: Will need to explicitly install the versions of .NET we need

Aaronontheweb (approved):
LGTM

PR #108, opened 2022-11-07
Fix inconsistent HOCON property naming and access, add MySql compat settings

Arkatufus:
## Changes

- The `persistenceId` property are named `PersistenceId` in one file but named `persistenceId` in another.
- Wrong property name were used when reading the HOCON settings
- Simplify HOCON by removing unnesscessary double quotes
- Add MySql compatibility settings
- Move default HOCON config loading from journal and snapshot plugin to Linq2DbPersistence class that extends IExtension

### DIFF
--- a/src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableColumnNames.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableColumnNames.cs
@@ -22,6 +22,9 @@ namespace Akka.Persistence.Sql.Linq2Db.Config
                 case "postgres":
                     colString = "postgres-compat-column-names";
                     break;
+                case "mysql":
+                    colString = "mysql-compat-column-names";
+                    break;
                 default:
                     colString = "column-names";
                     break;
@@ -32,7 +35,7 @@ namespace Akka.Persistence.Sql.Linq2Db.Config
                     ConfigurationFactory.ParseString(FallBack).GetConfig($"tables.journal.{colString}"));
             Ordering =       cfg.GetString("ordering","ordering");
             Deleted =        cfg.GetString("deleted","deleted");
-            PersistenceId =  cfg.GetString("PersistenceId", "persistence_id");
+            PersistenceId =  cfg.GetString("persistenceId", "persistence_id");
             SequenceNumber = cfg.GetString("sequenceNumber", "sequence_number");
             Created =        cfg.GetString("created", "created");
             Tags =           cfg.GetString("tags", "tags");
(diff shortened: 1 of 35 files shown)


src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableColumnNames.cs
```diff
@@ -32,7 +35,7 @@ public JournalTableColumnNames(Configuration.Config config)
                     ConfigurationFactory.ParseString(FallBack).GetConfig($"tables.journal.{colString}"));
             Ordering =       cfg.GetString("ordering","ordering");
             Deleted =        cfg.GetString("deleted","deleted");
-            PersistenceId =  cfg.GetString("PersistenceId", "persistence_id");
+            PersistenceId =  cfg.GetString("persistenceId", "persistence_id");
```
Aaronontheweb, line 36: LGTM

Aaronontheweb (approved):
LGTM - do we have tests to cover?

Arkatufus:
Just added all of the settings unit tests

src/Akka.Persistence.Sql.Linq2Db/Extension.cs
```diff
@@ -0,0 +1,60 @@
...
+        public static Configuration.Config DefaultConfiguration()
+        {
+            var journalConfig =  ConfigurationFactory.FromResource<Linq2DbWriteJournal>("Akka.Persistence.Sql.Linq2Db.persistence.conf");
+            var snapshotConfig = ConfigurationFactory.FromResource<Linq2DbSnapshotStore>("Akka.Persistence.Sql.Linq2Db.snapshot.conf");
+        
+            return journalConfig.WithFallback(snapshotConfig);
+        }    
+    
+        public readonly Configuration.Config DefaultJournalConfig;
+        public readonly Configuration.Config DefaultSnapshotConfig;
```
Arkatufus, line 32: The `DefaultJournalConfig` and `DefaultSnapshotConfig` are relative HOCON Config (`akka.persistence.journal.linq2db` and `akka.persistence.snapshot-store.linq2db`)

src/Akka.Persistence.Sql.Linq2Db/Extension.cs
```diff
@@ -0,0 +1,60 @@
...
+        public readonly Configuration.Config DefaultConfig;
```
Arkatufus, line 33: Default HOCON config is moved here

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -1,6 +1,5 @@
 ﻿akka.persistence {
   journal {
-    plugin = "akka.persistence.journal.linq2db"
```
Arkatufus, line 3: This should never be declared inside the reference configuration file.

src/Akka.Persistence.Sql.Linq2Db/snapshot.conf
```diff
@@ -1,73 +1,91 @@
 ﻿akka.persistence {
-  snapshot-store
-    {
-      plugin = "akka.persistence.snapshot-store.linq2db"
```
Arkatufus, line 4: This should never be declared inside the reference configuration file.

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -240,86 +269,11 @@
...
-          postgres-compat-metadata-column-names {
-            "persistenceId" = "persistence_id"
-            "sequenceNumber" = "sequence_nr"
-          }
-        }
+        tag-separator = ","
+      
+        dao = "Akka.Persistence.Sql.Linq2Db.Journal.DAO.ByteArrayJournalDao, Akka.Persistence.Sql.Linq2Db"
+      
+        tables.journal = ${akka.persistence.journal.linq2db.tables.journal}
```
Arkatufus, line 276: Since the query and journal `tables.journal` is the same, we can replace the whole block with a substitution

src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableColumnNames.cs
```diff
@@ -83,44 +84,5 @@ public override int GetHashCode()
             hashCode.Add(Manifest);
             return hashCode.ToHashCode();
         }
-        public static readonly string FallBack = @"tables.journal
```
Arkatufus, line 86: Baking hardwired HOCON config is a bad idea, this creates 4 copies of the same HOCON config in 4 different place

src/Akka.Persistence.Sql.Linq2Db/Config/MetadataTableColumnNames.cs
```diff
@@ -51,26 +52,5 @@ public override int GetHashCode()
         {
             return HashCode.Combine(PersistenceId, SequenceNumber);
         }
-
-        
-        
-        public static readonly string FallBack = @"tables.journal{
```
Arkatufus, line 57: Baking hardwired HOCON config is a bad idea, this creates 4 copies of the same HOCON config in 4 different place

src/Akka.Persistence.Sql.Linq2Db/Config/SnapshotTableColumnNames.cs
```diff
@@ -46,31 +48,5 @@ public override int GetHashCode()
             return HashCode.Combine(PersistenceId, SequenceNumber, Created,
                 Snapshot, Manifest, SerializerId);
         }
-
-        private static readonly string FallBack =
```
Arkatufus, line 50: Baking hardwired HOCON config is a bad idea, this creates 4 copies of the same HOCON config in 4 different place

src/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
```diff
@@ -47,17 +47,21 @@ public static DateTime FromUnixEpochMillis(in long unixEpochMillis)
 
     public class Linq2DbWriteJournal : AsyncWriteJournal
     {
+        [Obsolete(message: "Use Linq2DbPersistence.Get(ActorSystem).DefaultConfig instead")]
         public static Configuration.Config DefaultConfiguration =>
```
Aaronontheweb: Change this to `readonly` field rather than a property?

src/Akka.Persistence.Sql.Linq2Db/Extension.cs
```diff
@@ -0,0 +1,60 @@
...
+// -----------------------------------------------------------------------
+
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Persistence.Sql.Linq2Db.Journal;
+using Akka.Persistence.Sql.Linq2Db.Snapshot;
+
+namespace Akka.Persistence.Sql.Linq2Db
+{
+    public sealed class Linq2DbPersistence : IExtension
```
Aaronontheweb, line 14: What's the rationale for creating an `ActorSystemExternsion` here?

Aaronontheweb (changes requested):
Need some more context around these changes - looks like mostly configuration related stuff.

src/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
Arkatufus: Good idea

src/Akka.Persistence.Sql.Linq2Db/Extension.cs
Arkatufus: The biggest advantage is preloading the HOCON configuration. Parsing HOCON is a very expensive operation, for a given HOCON file, parsing should be done once and only once at the start of the plugin lifetime.

Arkatufus: Second is a centralized place to access the default configurations and this also follows the standardized flow we have on other persistence plugins

src/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
Arkatufus: Done

Aaronontheweb:
@Arkatufus looks like your tests are failing

Arkatufus:
Looks like racy tests

Aaronontheweb (approved):

PR #115, opened 2022-11-10
Fix several bugs

Arkatufus:
## Changes

- Tags should be formatted as ";{tag};"
- Highest journal sequence should be descending sorted
- Log should have exceptions as the first parameter

### DIFF
--- a/src/Akka.Persistence.Sql.Linq2Db.Tests/Settings/JournalConfigSpec.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db.Tests/Settings/JournalConfigSpec.cs
@@ -238,7 +238,7 @@ namespace Akka.Persistence.Sql.Linq2Db.Tests.Settings
             journal.DefaultSerializer.Should().BeNullOrEmpty();
 
             var pluginConfig = journal.PluginConfig;
-            pluginConfig.TagSeparator.Should().Be(",");
+            pluginConfig.TagSeparator.Should().Be(";");
             var daoType = Type.GetType(pluginConfig.Dao);
             daoType.Should().Be(typeof(ByteArrayJournalDao));
 
--- a/src/Akka.Persistence.Sql.Linq2Db.Tests/Settings/JournalConfigSpec.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db.Tests/Settings/JournalConfigSpec.cs
@@ -53,7 +53,7 @@ namespace Akka.Persistence.Sql.Linq2Db.Tests.Settings
             journal.GetBoolean("use-clone-connection").Should().BeFalse();
             journal.GetString("materializer-dispatcher", "invalid").Should().Be("akka.actor.default-dispatcher");
             journal.GetString("serializer", "invalid").Should().BeNullOrEmpty();
-            journal.GetString("tag-separator", "invalid").Should().Be(",");
+            journal.GetString("tag-separator", "invalid").Should().Be(";");
             journal.GetString("dao", "invalid").Should()
                 .Be("Akka.Persistence.Sql.Linq2Db.Journal.DAO.ByteArrayJournalDao, Akka.Persistence.Sql.Linq2Db");
 
--- a/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/ByteArrayJournalSerializer.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/ByteArrayJournalSerializer.cs
@@ -43,8 +43,7 @@ namespace Akka.Persistence.Sql.Linq2Db.Journal.DAO
                 return "";
             }
 
-            return tags.Aggregate((tl, tr) =>
-                tl + separator + tr);
+            return $"{separator}{string.Join(separator, tags)}{separator}";
         }
 
         protected override Try<JournalRow> Serialize(IPersistentRepresentation persistentRepr, IImmutableSet<string> tTags, long timeStamp = 0)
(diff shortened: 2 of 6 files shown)


src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/ByteArrayJournalSerializer.cs
```diff
@@ -43,8 +43,7 @@ public ByteArrayJournalSerializer(IProviderConfig<JournalTableConfig> journalCon
                 return "";
             }
 
-            return tags.Aggregate((tl, tr) =>
-                tl + separator + tr);
+            return $"{separator}{string.Join(separator, tags)}{separator}";
```
Arkatufus, line 46: Tags column should be formatted as ";{tag1};{tag2};{tag3};" with separator prefix and postfix

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -126,32 +126,22 @@ private static int MaxTake(long max)
             var maxTake = MaxTake(max);
             return AsyncSource<JournalRow>.FromEnumerable(new{separator,tag,offset,maxOffset,maxTake,_connectionFactory},
                     async(input)=>
-                {
+                    {
+                    var tag = $"{separator}{input.tag}{separator}";
```
Arkatufus, line 130: Tags were matched as "{separator}{tag}{separator}" so that we can be sure that the whole tag were matched

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -126,32 +126,22 @@ private static int MaxTake(long max)
...
                     {
                         return await conn.GetTable<JournalRow>()
-                            .Where<JournalRow>(r => r.tags.Contains(input.tag))
+                            .Where<JournalRow>(r => r.tags.Contains(tag))
                             .OrderBy(r => r.ordering)
                             .Where(r =>
                                 r.ordering > input.offset && r.ordering <= input.maxOffset)
                             .Take(input.maxTake).ToListAsync();
                     }
-                }).Via(perfectlyMatchTag(tag, separator))
```
Arkatufus, line 139: We do not need this check anymore because we are sure that the tag is matched perfectly

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -214,6 +204,7 @@ public async Task<long> MaxJournalSequenceAsync()
             {
                 return await db.GetTable<JournalRow>()
                     .Select<JournalRow, long>(r => r.ordering)
+                    .OrderByDescending(r => r)
```
Arkatufus, line 207: Sequence need to be explicitly sorted in descending order, there is no guarantee that all SQL server will automatically does this.

src/Akka.Persistence.Sql.Linq2Db/Query/JournalSequenceActor.cs
```diff
@@ -192,9 +192,7 @@ protected override void PreStart()
             {
                 if (t.IsFaulted)
                 {
-                    _log.Info(
-                        "Failed to recover fast, using event-by-event recovery instead",
-                        t.Exception);
+                    _log.Info(t.Exception, "Failed to recover fast, using event-by-event recovery instead");
```
Arkatufus, line 195: Exception should be the first parameter in a log method call

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -84,7 +84,7 @@
       # You may wish to provide a dedicated dispatcher instead
       materializer-dispatcher = "akka.actor.default-dispatcher"
       
-      tag-separator = ","
+      tag-separator = ";"
```
Arkatufus, line 87: The default tag separator has always been a semi colon in all SQL plugins

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -269,7 +269,7 @@
         #This setting will go away once https://github.com/linq2db/linq2db/issues/2466 is resolved
         use-clone-connection = false
 
-        tag-separator = ","
+        tag-separator = ";"
```
Arkatufus, line 272: The default tag separator has always been a semi colon in all SQL plugins

Arkatufus (commented):
Self-review. All bugs were found when attempting to implement tests based on historical data.

src/Akka.Persistence.Sql.Linq2Db.Tests/Settings/JournalConfigSpec.cs
```diff
@@ -53,7 +53,7 @@ public void DefaultJournalHoconConfigTest()
             journal.GetBoolean("use-clone-connection").Should().BeFalse();
             journal.GetString("materializer-dispatcher", "invalid").Should().Be("akka.actor.default-dispatcher");
             journal.GetString("serializer", "invalid").Should().BeNullOrEmpty();
-            journal.GetString("tag-separator", "invalid").Should().Be(",");
+            journal.GetString("tag-separator", "invalid").Should().Be(";");
```
Aaronontheweb, line 56: Nice catch

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
Aaronontheweb: Good catch

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
Aaronontheweb: Yep, want to be explicit about that in the query

src/Akka.Persistence.Sql.Linq2Db/Query/JournalSequenceActor.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM - great catches

PR #117, opened 2022-11-11
Add data compatibility checker

### LINKED ISSUE #3: Compatibility with Akka.Persistence.MySql
We do not have a configuration or a compatibility spec to work with Akka.Persistence.MySql.

In theory, a user would be able to manually configure the tables with `delete-compatibility-mode`, but we should provide easy configuration and ideally provide unit tests.

We should make sure such integration/implementation uses a MySql connector and Implementation packages that are non-encumbering (i.e. we should steer users away from adopting the Oracle MySqlConnector, but rather an MIT or Apache based one, consider MariaDB or similar instead of vanilla MySql)

Arkatufus:
Closes #3

## Changes

The new `Akka.Persistence.Linq2Db.Data.Compatibility` project is a test project that consumes docker images generated by the `Akka.Persistence.Sql.Exporter` repo.
The tests only covers Akka.Persistence and Akka.Cluster.Sharding with remember entity turned off. All remember entity data are present in the export data but are not checked.

Current tests:
- Full entity recovery and validation
- `CurrentEventsByTag` query test
- `CurrentPersistenceIds` query test, skipping system generated persistence Ids
- `CurrentEventsByPersistenceId` query test on all entities
- Journal events deletion using `logical-delete` turned on
- `CurrentEventsByTag` query test on journal table that has been deleted using `logical-delete` turned on


Arkatufus:
All of the unit tests are located inside the `DataCompatibility*` classes. 
Docker containers are loaded every time a test is being run, this adds overhead to the test. In a perfect world, a single instace of Docker container and Akka cluster can be used for all of the non-destructive tests. We might need to look into this if the test time became too unwieldy.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -0,0 +1,250 @@
...
+        
+        protected DataCompatibilitySpecBase(ITestOutputHelper output)
+        {
+            Output = output;
+            Fixture = new T();
+        }
+
+        private Config Config()
+        {
+            return (Config) $@"
```
Aaronontheweb, line 59: All of the pertinent compatibility settings are really from the derived `TestSettings` types - got it

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -0,0 +1,250 @@
...
+                    throw new TimeoutException("Failed to truncate all data within 10 seconds");
+                }
+            }
+            finally
+            {
+                cts.Dispose();
+            }
+        }
+        
+        protected static void ValidateState(string persistentId, StateSnapshot lastSnapshot, int count, int persisted)
```
Aaronontheweb, line 179: Where do the hard values like 36 come from?

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -0,0 +1,250 @@
...
+            persisted.Should().Be(36, "Entity {0} should have persisted 36 events", persistentId);
+            lastSnapshot.Persisted.Should().Be(24, "Entity {0} last snapshot should have persisted 24 events", persistentId);
+
+            var baseValue = int.Parse(persistentId) * 3;
+            var roundTotal = (baseValue * 3 + 3) * 4;
+            lastSnapshot.Total.Should().Be(roundTotal * 2, "Entity {0} last snapshot total should be {1}", persistentId, roundTotal * 2);
+            count.Should().Be(roundTotal * 3, "Entity {0} total should be {1}", persistentId, roundTotal * 3);
+        }
+
+        protected static async Task ValidateTags(ActorSystem system)
```
Aaronontheweb: I'd also explain in a comment here where the hard numbers come from

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpec.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+        
+        [Fact(DisplayName = "Linq2Db should recover data created using other persistence plugins")]
+        public async Task RecoveryValidationTest()
+        {
+            await ValidateRecovery(TestCluster!.ShardRegions[0]);
+        }
+
+        [Fact(DisplayName = "Linq2Db should be able to retrieve persistence ids on data created by other persistence plugin")]
+        public async Task PersistenceIdQueryValidationTest()
```
Aaronontheweb, line 42: LGTM - but should add a comment explaining how many IDs are stored in the database (be explicit)

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/MySql/MySqlSpecSettings.cs
```diff
@@ -0,0 +1,29 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="MySqlSpecSettings.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
```
Aaronontheweb: I'd remove the Lightbend attribution - none of this is a port

Aaronontheweb (approved):
Overall, LGTM

Arkatufus:
### NOTE
~~Will not pass test until https://github.com/akkadotnet/Akka.Persistence.Linq2Db/pull/121 is merged~~

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpec.cs
to11mtm: On a related note, we should make sure that we are handling 'loop-overs' on these (or other persistence.query) tests, i.e. make sure we are doing enough events to where multiple batches to store are required.

Aaronontheweb: Could you help me understand this better? Running repeated queries over the same data set or running multiple queries concurrently?

to11mtm: I could have worded that better.

Uhh, long story short, when we do reads on the DB, we chunk them. I'd have to look to find which number it is (I haven't ever really used the query bits) but if I had to guess it's the `replay-batch-size`. 

tl;dr- if we are aiming to handle compatibility as well as general coverage I would encourage making sure there's enough events that the query stream does at least two hits to the DB to fetch them all.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpec.cs
```diff
@@ -0,0 +1,133 @@
...
+            var readJournal = PersistenceQuery.Get(system)
+                .ReadJournalFor<Linq2DbReadJournal>(Linq2DbReadJournal.Identifier);
+
+            foreach (var id in Enumerable.Range(0, 100))
+            {
+                var baseValue = id * 3;
+                var roundTotal = (baseValue * 3 + 3) * 4;
+
+                var events = await readJournal.CurrentEventsByPersistenceId(id.ToString(), 0, long.MaxValue)
+                    .RunAsAsyncEnumerable(system.Materializer()).ToListAsync();
```
to11mtm, line 68: Happy that `RunAsAsyncEnumerable` is useful, although It does still scare me sometimes 😅

Arkatufus: Its pretty safe, as long as we know that the event source is not an infinite source

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpec.cs
Arkatufus: Can we limit the batch fetch to less than the data in the database? The data is a known variable, each entity would have 12 events in the journal to be replayed since the last snapshot.
So if we limit the batch fetch to 6, that would force Linq2Db to do 2 chunks of fetches?

Aaronontheweb: Yep, making the batch size artificially smaller for testing is a good solution here, to ensure that pagination works. Let's give that a try.

Arkatufus: @to11mtm 
Is this the setting you're referring to?
https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/d0cfde22388c0a7f5ba33f01929b1d9083e23d10/src/Akka.Persistence.Sql.Linq2Db/persistence.conf#L49

or is it this one?
https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/d0cfde22388c0a7f5ba33f01929b1d9083e23d10/src/Akka.Persistence.Sql.Linq2Db/persistence.conf#L60

also this one?
https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/d0cfde22388c0a7f5ba33f01929b1d9083e23d10/src/Akka.Persistence.Sql.Linq2Db/persistence.conf#L252

Would it be safe to set `batch-size` to 3 and `db-round-trip-max-batch-size` and `replay-batch-size` to 6?

Arkatufus: @to11mtm Do you have any unit test suggestion to test these?

to11mtm: @Arkatufus I will be honest and say I don't know which setting is correct off the top of my head, but you could probably turn all of them down for the sake of the compatibility tests.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilityLogicalDeleteSpec.cs
```diff
@@ -0,0 +1,78 @@
...
+    public abstract class DataCompatibilityLogicalDeleteSpec<T> : DataCompatibilitySpecBase<T> where T : ITestContainer, new()
+    {
+        protected DataCompatibilityLogicalDeleteSpec(ITestOutputHelper output): base(output)
+        {
+        }
+
+        protected override void Setup(AkkaConfigurationBuilder builder, IServiceProvider provider)
+        {
+            builder.AddHocon((Config)@"
+akka.persistence.journal.linq2db.logical-delete = true
```
Aaronontheweb, line 33: I know there's another open issue for this, but we should not include this option IMHO - a delete is a delete.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilityLogicalDeleteSpec.cs
```diff
@@ -0,0 +1,78 @@
...
+akka.persistence.query.journal.linq2db.include-logically-deleted = true", HoconAddMode.Prepend);
```
Aaronontheweb, line 34: I don't think this option should exist either.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -0,0 +1,260 @@
...
+	journal {{
+		plugin = ""akka.persistence.journal.linq2db""
+		linq2db {{
+			connection-string = ""{Fixture.ConnectionString}""
+			provider-name = {Settings.ProviderName}
+			table-compatibility-mode = {Settings.CompatibilityMode}
+
+            # Testing for https://github.com/akkadotnet/Akka.Persistence.Linq2Db/pull/117#discussion_r1027345449
+            batch-size = 3
+            db-round-trip-max-batch-size = 6
```
Aaronontheweb, line 70: Made these values smaller in order to guarantee that batching in queries works correctly

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -0,0 +1,260 @@
...
+	}}
+
+	query.journal.linq2db {{
+		connection-string = ""{Fixture.ConnectionString}""
+		provider-name = {Settings.ProviderName}
+		table-compatibility-mode = {Settings.CompatibilityMode}
+
+        # Testing for https://github.com/akkadotnet/Akka.Persistence.Linq2Db/pull/117#discussion_r1027345449
+        batch-size = 3
+        replay-batch-size = 6
```
Aaronontheweb, line 89: Same as above but for Akka.Persistence.Query specifically

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/xunit.runner.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://xunit.github.io/schema/current/xunit.runner.schema.json",
+  "longRunningTestSeconds": 60,
+  "parallelizeAssembly": false,
+  "parallelizeTestCollections": false
```
Aaronontheweb, line 5: LGTM

Aaronontheweb (approved):
LGTM

PR #121, opened 2022-11-11
Fix read journal, missing `_includeDeleted` checks

Arkatufus:
Prerequisite for #117

Persistence query returned wrong number of events because it also includes soft deleted entries.

### DIFF
--- a/src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
@@ -39,16 +39,12 @@ namespace Akka.Persistence.Sql.Linq2Db.Query.Dao
             _deserializeFlow = _serializer.DeserializeFlow();
         }
 
-        protected IQueryable<JournalRow> BaseQuery(DataConnection connection)
-        {
-            return connection.GetTable<JournalRow>()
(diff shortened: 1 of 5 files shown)


to11mtm (approved):

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -171,7 +167,7 @@ public async Task<long> MaxJournalSequenceAsync()
         {
             await using var db = ConnectionFactory.GetConnection();
             
-            return await db.GetTable<JournalRow>()
+            return await BaseQueryStatic(db, _includeDeleted)
```
to11mtm, line 170: So, I'm realizing now (After approval) that this line should not have been changed;

Here are the queries from AFTER 'include-deleted' support was pulled from persistence-jdbc (But before license change! ;))

https://github.com/akka/akka-persistence-jdbc/blob/933ece9ae13c4ae9934ffc01e67ea477cfd8d85c/core/src/main/scala/akka/persistence/jdbc/query/dao/ReadJournalQueries.scala 

The max journal 'sequence' (Ordering) does -not- have a filter on isDeleted. If I had to guess, including deleted in Max Ordering -probably- improves some edge cases around `JournalSequenceActor`, namely in cases where all of the entries in the journal are soft-delete records, (If that happened, there would be a delay on seeing new events, as the Journal Sequence actor would take time to look for the sequence gap.)

PR #130, opened 2022-12-13
Refactor compatibility settings to be more generic

Arkatufus:
## Changes

- Refactor the table compatibility configuration so that all of the database related mappings are contained within a single block so that it is easier to read at a glance.
- Use the original names of the persistent plugins (ie. "sql-server" and "postgresql") so it is more relateable to the users.
- Compatibility table names are embedded inside the config block, no need to remember to change table names to make compatibility work, less headache and less error prone for users who migrated from vanilla persistence.sql plugins.
- Move auto initialize to the plugin level block and use the standardized `auto-initialize` setting name.
- Set auto initialize default value to false, just like the other plugins in the ecosystem.
- Less confusing code to add additional tables, when needed.
- Removed magic strings, less things to break on future updates.


src/Akka.Persistence.Linq2Db.Benchmark.DockerTests/Docker/Linq2Db/DockerLinq2DbPostgreSQLJournalPerfSpec.cs
```diff
@@ -30,10 +30,12 @@ class = ""{typeof(Linq2DbWriteJournal).AssemblyQualifiedName}""
             connection-string = ""{connString}""
             provider-name = ""{ProviderName.PostgreSQL95}""
             use-clone-connection = true
-            tables.journal {{ 
-                auto-init = true
-                warn-on-auto-init-fail = false
-                table-name = ""testPerfTable"" 
+            auto-initialize = true
```
Aaronontheweb, line 33: This brings the setting back to where it usually is for all other Akka.Persistence implementations (in the plugin's `.journal` HOCON)

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -89,102 +101,156 @@
...
+          table-name = "tags"
+          columns {
+            ordering-id = ordering_id
+            tag-value = tag
+            writer-uuid = writer_uuid
+          }
+        }
+      }
+      
+      # Akka.Persistence.SqlServer compatibility table name and column name mapping
```
Aaronontheweb, line 134: So if I'm interpreting this correctly: you wanted to make the relationships between column mappings and their respective database engines more explicit and clearer reason about by breaking them up into their own independent / nestable sections, which all tie back into ` table-mapping = [value]` - because in theory, you could replace `value` with something entirely custom and that would work - say, for instance, Oracle support?

src/Akka.Persistence.Sql.Linq2Db/snapshot.conf
```diff
@@ -30,62 +31,90 @@
...
+      auto-initialize = false
+      
+      #if true, a warning will be logged
+      #if auto-init of tables fails.
+      #set to false if you don't want this warning logged
+      #perhaps if running CI tests or similar.
+      warn-on-auto-init-fail = true
       
-      tables.snapshot {
+      default {
```
Aaronontheweb, line 44: Same idea as the journal - got it (provided that my assumptions in my comment on the journal are correct - please let me know @Arkatufus )

Aaronontheweb (commented):
So this is actually a much narrower change than it looks given the number of affected files, which is because this organization of the configuration is somewhat viral.

It looks to me like overall the goal here was to make the table mappings more configurable / customizable by changing the structure of the relationship expressed in the HOCON (i.e. linking to a named section versus having hard-coded keys for specific databases.)

LGTM but I'd like to have @Arkatufus answer my one question I left in the comments.

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
Arkatufus: Exactly. The table mapping is completely decoupled now, it is generic enough that people can add their own entries to an SQL server that we don't even know. It should be safe enough as long as these 3rd party plugins follows our standard table schema.

Arkatufus: For clarity, this is the database defining HOCON block:
```hocon
database-name {
    schema-name = null # schema for the database
    
    journal {
        table-name = journal    # event journal table name mapping
        columns {               # column name mapping block
            ordering = ordering # column name mapping
            deleted = deleted
            persistenceId = persistence_id
            sequenceNumber = sequence_number
            created = created
            tags = tags
            message = message
            identifier = identifier
            manifest = manifest
        }
    }
    
    metadata { # same concept for the metadata table
        table-name = journal_metadata
        columns {
            persistenceId = persistenceId
            sequenceNumber = sequenceNr
        }
    }
    
    tag { # same concept for the future tag table
        table-name = tags
        columns {
            ordering-id = ordering_id
            tag-value = tag
            writer-uuid = writer_uuid
        }
    }
}
```

src/Akka.Persistence.Linq2Db.Compatibility.DockerTests/SqlServer/SqlServerCompatibilitySpecConfig.cs
```diff
@@ -31,15 +31,13 @@ class = ""Akka.Persistence.SqlServer.Snapshot.SqlServerSnapshotStore, Akka.Persi
...
             connection-string = ""{DbUtils.ConnectionString}""
-            provider-name = """ + LinqToDB.ProviderName.SqlServer2017 + $@"""
-            table-compatibility-mode = sqlserver
-            tables {{
-                snapshot {{ 
-                    auto-init = true
-                    warn-on-auto-init-fail = false
-                    table-name = ""{tableName}""    
+            provider-name = ""{LinqToDB.ProviderName.SqlServer2017}""
+            table-mapping = sql-server
```
to11mtm, line 36: My biggest concern with this is that `table-mapping` is somewhat inconsistent with overall *intent* of this being a compatibility mapping for the old plugin.

i.e. if there is no `table-compatibility-mode` set, we don't use `metadata` at all for deletes or maxSeq checks (which is of course faster.)

src/Akka.Persistence.Linq2Db.Compatibility.DockerTests/SqlServer/SqlServerCompatibilitySpecConfig.cs
```diff
@@ -31,15 +31,13 @@ class = ""Akka.Persistence.SqlServer.Snapshot.SqlServerSnapshotStore, Akka.Persi
         linq2db {{
             class = ""{typeof(Linq2DbSnapshotStore).AssemblyQualifiedName}""
             plugin-dispatcher = ""akka.persistence.dispatchers.default-plugin-dispatcher""
-            #plugin-dispatcher = ""akka.actor.default-dispatcher""
             connection-string = ""{DbUtils.ConnectionString}""
-            provider-name = """ + LinqToDB.ProviderName.SqlServer2017 + $@"""
-            table-compatibility-mode = sqlserver
-            tables {{
```
to11mtm, line 38: FWIW, this 'tables' structure in the HOCON was borrowed from akka-persistence-jdbc. I don't think there is 'much' we don't have compared to the last 'pre-lockdown' commit they have, but it -may- have a small impact on those remaining bits.

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/SqlServer/SqlServerSpecSettings.cs
```diff
@@ -14,15 +14,9 @@ private SqlServerSpecSettings()
...
-        public override string JournalTableName => "EventJournal";
-        
-        public override string JournalMetadataTableName => "Metadata";
-        
-        public override string SnapshotTableName => "SnapshotStore";
-        
         public override string ProviderName => LinqToDB.ProviderName.SqlServer2017;
         
-        public override string CompatibilityMode => "sqlserver";
+        public override string TableMapping => "sql-server";
```
to11mtm, line 19: Since it's relatively low effort in the hocon->class wireup, is it possible to still *allow* the old versions to avoid breaking them? (looking out for some folks who I know use this package that might run into problems if they don't catch it.)

src/Akka.Persistence.Sql.Linq2Db/Config/EventJournalTableConfig.cs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override bool Equals(object obj)
+    {
+        return ReferenceEquals(this, obj) || obj is EventJournalTableConfig other && Equals(other);
+    }
+
+    public override int GetHashCode()
+    {
+        return HashCode.Combine(Name, ColumnNames);
```
to11mtm, line 37: stupid question; Does `HashCode.Combine` properly handle the fact `ColumnNames` is an object?

src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
```diff
@@ -0,0 +1,19 @@
...
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+
+namespace Akka.Persistence.Sql.Linq2Db.Config;
+
+public sealed class TagTableColumnNames
```
to11mtm: Should this be here? looks premature.

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
to11mtm: > So if I'm interpreting this correctly: you wanted to make the relationships between column mappings and their respective database engines more explicit and clearer reason about by breaking them up into their own independent / nestable sections, which all tie back into ` table-mapping = [value]` - because in theory, you could replace `value` with something entirely custom and that would work - say, for instance, Oracle support?

Well, it -might- be a little more involved than that.

the `table-mapping` is only sets up the compatibility mapping for old tables (otherwise, we use the column names more or less specced by persistence-jdbc). `provider-name` is what actually sets up the per-DB mapping.

I make that disclaimer, because some database providers may or may not require additional mapping to deal with compatibility (main example, [SQLite and PG snapshot timestamps](https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/4e48a79e0e6a6c1ac6fbbd29355966e4c4a758d8/src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs#L102-L107)) 

FWIW most mapping changes should be minor, (most typically around timestamp handling on snapshots if I had to guess.)

to11mtm (approved):

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/SqlServer/SqlServerSpecSettings.cs
Arkatufus: Yes, since its just a database name mapping, that is easily doable using hocon substitution

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
Arkatufus: You have a point, the table column data type themselves are not standardized, some would need adapters to make them work.

Arkatufus: If I remember correctly, those adapters are actually applied based on the provider name, and not by the compatibility table name, so these changes would not affect those.

src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
Arkatufus: Yeah, those snuck in from another PR. I'll remove those.

Aaronontheweb: So we need to remove this - were there other compat changes that needed to be introduced to this PR to address the rest of @to11mtm 's concerns?

src/Akka.Persistence.Sql.Linq2Db/Config/JournalTableConfig.cs
```diff
@@ -3,40 +3,42 @@
...
-            MetadataColumnNames = new MetadataTableColumnNames(config);
-            TableName = localCfg.GetString("table-name", "journal");
-            MetadataTableName = localCfg.GetString("metadata-table-name", "journal_metadata");
-            SchemaName = localCfg.GetString("schema-name", null);
-            AutoInitialize = localCfg.GetBoolean("auto-init", false);
-            WarnOnAutoInitializeFail = localCfg.GetBoolean("warn-on-auto-init-fail", true);
+            // backward compatibility
+            var compat = config.GetString("table-compatibility-mode");
+            if (compat != null)
+                mappingPath = compat;
```
Arkatufus, line 20: @to11mtm added the compatibility code here

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -89,102 +101,151 @@
...
+            persistence-id = persistence_id
+            sequence-number = sequence_nr
+          }
         }
-        postgres-compat-metadata-column-names {
-          persistenceId = persistence_id
-          sequenceNumber = sequence_nr
+      }
+
+      postgres = ${akka.persistence.journal.linq2db.postgresql} # backward compat naming
```
Arkatufus, line 218: @to11mtm backward compat config here

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
```diff
@@ -89,102 +101,151 @@
...
-          persistenceId = persistence_id
-          sequenceNumber = sequence_nr
-          created = created_at
-          tags = tags
-          message = payload
-          identifier = serializer_id
-          manifest = manifest
+      }
+      
+      sqlserver = ${akka.persistence.journal.linq2db.sql-server} # backward compat naming
```
Arkatufus, line 161: @to11mtm backward compat config here

src/Akka.Persistence.Sql.Linq2Db/persistence.conf
to11mtm: Correct. The above was mostly clarification on work for future DB support (i.e. we may or may not need to add provider specific mappings in addition to this config.)

PR #132, opened 2022-12-21
Refactor compatibility unit tests to use TestKit

Arkatufus:
Fixes unit test problem where new actor names kept colliding with dead ones.

## Changes

Refactor compatibility unit tests to use TestKit


Aaronontheweb (approved):
LGTM

PR #135, opened 2023-02-03
Add batched migration script

Arkatufus:
## Changes
Add batched migration script


scripts/SqlServer/Batched/1_Migration_Setup.sql
```diff
@@ -0,0 +1,69 @@
...
+    WHERE
+        ej.[Tags] IS NOT NULL
+      AND LEN(ej.[Tags]) > 0
+      AND ej.[Ordering] NOT IN (SELECT t.[ordering_id] FROM [dbo].[tags] t);
+        
+    WHILE @from_id <= @max_id
+    BEGIN
+        BEGIN TRAN;
+        INSERT INTO [dbo].[tags]([ordering_id], [tag])
+        SELECT * FROM (SELECT records.[Ordering], cross_product.[items] FROM (
```
to11mtm: Would prefer specific columns over select * here

scripts/SqlServer/Batched/1_Migration_Setup.sql
```diff
@@ -0,0 +1,69 @@
...
+            SELECT *
+            FROM [dbo].[EventJournal] AS ej
+            WHERE ej.[Ordering] >= @from_id AND ej.[Ordering] <= @from_id + 1000
+        ) AS records CROSS APPLY [dbo].[Split](records.Tags, ';') cross_product) AS s([ordering_id], [tag])
+        WHERE NOT EXISTS (
+                SELECT * FROM [dbo].[tags] t WITH (updlock)
```
to11mtm, line 64: You can probably `SELECT 1` instead of `SELECT *` here.

to11mtm (commented):
@Arkatufus I'm looking at these with the mindset that they are expected to be modified for a specific use case.

That aside,

- We should give our function names unique names if we are using `CREATE OR ALTER`, especially in the case of functions named things like `Split` which may have an existing declaration in a Prod DB.
- We should have documentation that makes it abundantly clear to users that these scripts must be modified for their specific table configuration(s).

Arkatufus:
@to11mtm makes sense, I'll try to incorporate those changes

scripts/SqlServer/Batched/1_Migration_Setup.sql
Arkatufus: The `SELECT *` is actually nescessary here, the specific columns are defined inside the inner `SELECT`

Aaronontheweb:
@Arkatufus needs a rebase

Arkatufus:
Rebased to recent dev branch in #168

PR #138, opened 2023-02-03
Change default journal read and write mode to tag table

Arkatufus:
* Change default journal read and write mode
* Fix DAO logic for tag table

### DIFF
--- a/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
@@ -15,7 +15,7 @@ using Akka.Persistence.Sql.Linq2Db.Serialization;
 using Akka.Streams;
 using Akka.Streams.Dsl;
 using LanguageExt;
-using LinqToDB;using LinqToDB.Common;
+using LinqToDB;
 using LinqToDB.Data;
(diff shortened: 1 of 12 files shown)


Aaronontheweb (approved):
LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -156,115 +154,73 @@ private async Task QueueWriteJournalRows(Seq<JournalRow> xs)
...
-            }
-        }
-
-        private async Task HandleTagTableInsert(Seq<JournalRow> xs)
-        {
-            await using var db = ConnectionFactory.GetConnection();
-            await using var tx = await db.BeginTransactionAsync(IsolationLevel.ReadCommitted);
-            try
-            {
-                await ConsumeSequenceForTagInsert(xs, db);
```
to11mtm, line 267: So, the purpose of `ConsumeSequenceForTagInsert` was as follows:

Because we are batching all of our writes, but bulk copy is -much- faster than row-by-row, the idea was that in cases of:

`ItemWithTags ItemWithTags ItemNoTags ItemNoTags ItemNoTags ItemNoTags ItemWithTags ItemNoTags` etc... we would still be able to take advantage of bulk-copy for the no-tag cases, and at least lessen the write throughput penalty.

PR #141, opened 2023-02-07
Add original DB insertion optimization

### LINKED ISSUE #140: Need to reintroduce row insertion optimization
So, the purpose of `ConsumeSequenceForTagInsert` was as follows:

Because we are batching all of our writes, but bulk copy is -much- faster than row-by-row, the idea was that in cases of:

`ItemWithTags ItemWithTags ItemNoTags ItemNoTags ItemNoTags ItemNoTags ItemWithTags ItemNoTags` etc... we would still be able to take advantage of bulk-copy for the no-tag cases, and at least lessen the write throughput penalty.

_Originally posted by @to11mtm in https://github.com/akkadotnet/Akka.Persistence.Linq2Db/pull/138#discussion_r1098999591_

Arkatufus:
Fixes #140

## Changes
Reintroduce optimization


Arkatufus:
@to11mtm Can you give this PR a review and see if I correctly re-implement your optimization?

Aaronontheweb:
@to11mtm can this be merged?

PR #145, opened 2023-02-09
Fix TagTable delete and queries operations

Arkatufus:
## Changes

* Fix queries:
  * Always populate tags in tag table queries
  * Simplify EventsByTag query
* Denormalize TagTable by storing persistence id and sequence number
* Add TagTable row deletion

### DIFF
--- a/src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
+++ b/src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
@@ -9,6 +9,8 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
+using Akka.Cluster.Sharding;
+using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
 using Akka.Hosting;
 using Akka.Persistence.Linq2Db.Data.Compatibility.Tests.Internal;
@@ -48,9 +50,9 @@ namespace Akka.Persistence.Linq2Db.Data.Compatibility.Tests
             Fixture = new T();
         }
 
-        private Config Config()
+        protected Config Config()
         {
-            return (Config) $@"
+            return ((Config) $@"
 akka.persistence {{
 	journal {{
 		plugin = ""akka.persistence.journal.linq2db""
@@ -100,26 +102,29 @@ akka.persistence {{
             }}" : "")}
 		}}
 	}}
-}}";
+}}")
+                .WithFallback(Linq2DbPersistence.DefaultConfiguration)
+                .WithFallback(ClusterSharding.DefaultConfig())
+                .WithFallback(ClusterSingletonManager.DefaultConfig());
         }
 
-        protected virtual void CustomSetup(AkkaConfigurationBuilder builder, IServiceProvider provider)
+        protected virtual Task InitializeTestAsync()
         {
+            return Task.CompletedTask;
         }
 
         protected abstract void Setup(AkkaConfigurationBuilder builder, IServiceProvider provider);
         
         private void InternalSetup(AkkaConfigurationBuilder builder, IServiceProvider provider)
         {
-            builder.AddHocon(
-                Config().WithFallback(Linq2DbPersistence.DefaultConfiguration), 
-                HoconAddMode.Prepend);
+            builder.AddHocon(Config(), HoconAddMode.Prepend);
             Setup(builder, provider);
         }
 
         public async Task InitializeAsync()
         {
             await Fixture.InitializeAsync();
+            await InitializeTestAsync();
             TestCluster = new TestCluster(InternalSetup, "linq2db", Output);
             await TestCluster.StartAsync();
         }
@@ -134,35 +139,21 @@ akka.persistence {{
         protected async Task TruncateEventsToLastSnapshot()
         {
             var region = TestCluster!.ShardRegions[0];
-            var cts = new CancellationTokenSource(10.Seconds());
-            try
+            using var cts = new CancellationTokenSource(10.Seconds());
+            foreach (var id in Enumerable.Range(0, 100))
             {
-                var tasks = Enumerable.Range(0, 100).Select(id =>
-                    region.Ask<(string, StateSnapshot?)>(new Truncate(id), cts.Token)).ToList();
-                while (tasks.Count > 0)
-                {
-                    var task = await Task.WhenAny(tasks);
-                    if (cts.Token.IsCancellationRequested)
-                        break;
-
-                    tasks.Remove(task);
-                    var (id, lastSnapshot) = task.Result;
-                    if (lastSnapshot is { })
-                        Output.WriteLine(
-                            $"{id} data truncated. " +
-                            $"Snapshot: [Total: {lastSnapshot.Total}, Persisted: {lastSnapshot.Persisted}]");
-                    else
-                        throw new XunitException($"Failed to truncate events for entity {id}");
-                }
-
-                if (cts.IsCancellationRequested)
-                {
-                    throw new TimeoutException("Failed to truncate all data within 10 seconds");
-                }
+                var (_, lastSnapshot) = await region.Ask<(string, StateSnapshot?)>(new Truncate(id), cts.Token);
+                if (lastSnapshot is { })
+                    Output.WriteLine(
+                        $"{id} data truncated. " +
+                        $"Snapshot: [Total: {lastSnapshot.Total}, Persisted: {lastSnapshot.Persisted}]");
+                else
+                    throw new XunitException($"Failed to truncate events for entity {id}");
             }
-            finally
+
+            if (cts.IsCancellationRequested)
             {
-                cts.Dispose();
+                throw new TimeoutException("Failed to truncate all data within 10 seconds");
             }
         }
         
--- a/src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
+++ b/src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
@@ -16,10 +16,14 @@ public class TagTableColumnNames: IEquatable<TagTableColumnNames>
         var cfg = config.GetConfig("columns");
         OrderingId = cfg.GetString("ordering-id", "ordering_id");
         Tag = cfg.GetString("tag-value", "tag");
+        SequenceNumber = cfg.GetString("sequence-nr", "sequence_nr");
+        PersistenceId = cfg.GetString("persistence-id", "persistence_id");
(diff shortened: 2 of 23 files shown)


src/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalTagRow.cs
```diff
@@ -9,7 +9,8 @@ namespace Akka.Persistence.Sql.Linq2Db.Journal.Types;
 
 public sealed class JournalTagRow
 {
-    public long JournalOrderingId { get; set; }
-    
+    public long SequenceNumber { get; set; }
+    public string PersistenceId { get; set; }
+    public long OrderingId { get; set; }
```
Arkatufus, line 14: TagTable rows are expanded to include the event persistence id and sequence number. That way we can delete TagTable entries by following the same arguments as journal events deletion arguments (by using persistence id and sequence number)

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -134,35 +139,21 @@ public async Task DisposeAsync()
         protected async Task TruncateEventsToLastSnapshot()
         {
             var region = TestCluster!.ShardRegions[0];
-            var cts = new CancellationTokenSource(10.Seconds());
-            try
+            using var cts = new CancellationTokenSource(10.Seconds());
+            foreach (var id in Enumerable.Range(0, 100))
```
Arkatufus, line 143: * Use `using` instead of `try...finally`
* Delete single event at a time instead of batching all deletes, problematic with MySql and PostgreSql connection pool

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -335,6 +337,13 @@ await db.GetTable<JournalMetaData>()
                         .DeleteAsync();
                 }
 
+                if (JournalConfig.TableConfig.TagWriteMode != TagWriteMode.Csv)
+                {
+                    await db.GetTable<JournalTagRow>()
+                        .Where(r => r.SequenceNumber <= maxSequenceNr && r.PersistenceId == persistenceId)
+                        .DeleteAsync();
```
Arkatufus, line 344: TagTable row deletion, simplified because we don't have to cache the ordering numbers that has been deleted

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -240,14 +179,21 @@ private static async Task AddTagDataFromTagTable(List<JournalRow> toAdd, DataCon
...
+                            r.PersistenceId == state.persistenceId
                             && r.SequenceNumber >= state.fromSequenceNr
                             && r.SequenceNumber <= state.toSequenceNr
                             && r.Deleted == false)
                         .OrderBy(r => r.SequenceNumber)
-                        .Take(state.toTake).ToListAsync())
+                        .Take(state.toTake).ToListAsync();
+
+                    if (_readJournalConfig.PluginConfig.TagReadMode == TagReadMode.TagTable)
+                        await AddTagDataFromTagTable(mainRows, dc);
```
Arkatufus, line 191: Make sure that tags are populated on every query

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
```diff
@@ -134,35 +139,21 @@ public async Task DisposeAsync()
...
-                    else
-                        throw new XunitException($"Failed to truncate events for entity {id}");
-                }
-
-                if (cts.IsCancellationRequested)
-                {
-                    throw new TimeoutException("Failed to truncate all data within 10 seconds");
-                }
+                var (_, lastSnapshot) = await region.Ask<(string, StateSnapshot?)>(new Truncate(id), cts.Token);
+                if (lastSnapshot is { })
```
Aaronontheweb, line 146: I think I might be too much of a dinosaur, but what does this pattern match mean? Not null?

src/Akka.Persistence.Sql.Linq2Db/Config/TagTableColumnNames.cs
```diff
@@ -15,10 +15,14 @@ public TagTableColumnNames(Configuration.Config config)
         var cfg = config.GetConfig("columns");
         OrderingId = cfg.GetString("ordering-id", "ordering_id");
         Tag = cfg.GetString("tag-value", "tag");
+        SequenceNumber = cfg.GetString("sequence-nr", "sequence_nr");
+        PersistenceId = cfg.GetString("persistence-id", "persistence_id");
```
Aaronontheweb, line 19: LGTM

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -199,13 +194,33 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
                     .IsColumn().IsNullable(false)
                     .HasLength(64)
                     .IsPrimaryKey()
-                    .Member(r => r.JournalOrderingId).HasColumnName(tagColumns.OrderingId)
-                    .IsColumn().IsPrimaryKey();
+                    .Member(r => r.SequenceNumber).HasColumnName(tagColumns.SequenceNumber)
+                    .IsColumn().IsNullable(false)
+                    .Member(r => r.PersistenceId).HasColumnName(tagColumns.PersistenceId)
+                    .HasLength(255)
```
Aaronontheweb, line 200: Asking because I don't know the answer: where did the 255 character constraint come from for PersistentId?

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -199,13 +194,33 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfigu
...
+                    .IsColumn().IsNullable(false).IsPrimaryKey();
                 
                 if (config.ProviderName.ToLower().Contains("sqlite"))
```
Aaronontheweb, line 203: Why is the branch needed for Sqlite?

src/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
Aaronontheweb: LGTM

src/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalTagRow.cs
Aaronontheweb: Makes sense - remember discussing this a while ago. This is something I wish we could have done with Akka.Persistence.Azure but the query model doesn't easily allow for it.

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
```diff
@@ -141,95 +140,35 @@ private static async Task AddTagDataFromTagTable(List<JournalRow> toAdd, DataCon
...
+                            async input =>
+                            {
+                                await using var conn = input.ConnectionFactory.GetConnection();
+                                var journalTable = conn.GetTable<JournalRow>();
+                                var tagTable = conn.GetTable<JournalTagRow>();
+                                var query =
+                                    from r in journalTable
+                                    from lp in tagTable.Where(jtr => 
+                                        jtr.OrderingId == r.Ordering 
+                                        && jtr.PersistenceId == r.PersistenceId).DefaultIfEmpty()
```
Aaronontheweb: Why do we need to join on `PeristenceId` here? Is this _just_ for deletes?

Aaronontheweb (commented):
Left some questions

src/Akka.Persistence.Linq2Db.Data.Compatibility.Tests/DataCompatibilitySpecBase.cs
Arkatufus: Yes, `myVar is { }` is the new pattern matching way of saying that `myVar is not null` or `myVar != null`

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
Arkatufus: Its the default value for all `PersistenceId` column for all SQL extensions

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
Arkatufus: Its to prevent linq2db from storing the numbers in "REAL" floating point data type, explicitly state that we want the column to have "INTEGER" data type.

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
Arkatufus: You're right, let me fix that

src/Akka.Persistence.Sql.Linq2Db/Db/AkkaPersistenceDataConnectionFactory.cs
Aaronontheweb: Figures, ok

src/Akka.Persistence.Sql.Linq2Db/Query/Dao/BaseByteReadArrayJournalDAO.cs
Aaronontheweb: @Arkatufus we should probably have a unit test that would have caught this on the read-side - is there one sitting in an unmerged PR somewhere?

Arkatufus: Not yet, unit testing will be another PR after this

Aaronontheweb (approved):
LGTM

PR #154, opened 2023-03-10
feat: move to centralized package management

CumpsD:
Converted the solution to centralized package management according to the way forward for all Akka.net repos.


CumpsD:
After this PR is merged I'll make some more for:
* Rename to Akka.Persistence.Sql
* Rename postgres to postgresql
* Rename `SqLite` to `Sqlite`
* Make style consistent
* Upgrade versions

Aaronontheweb:
LGTM

PR #168, opened 2023-03-21
[REBASE #135] Add batched migration script

Arkatufus:
Rebased from #135

## Changes
Add batched migration script


Aaronontheweb:
@Arkatufus we're going to test this separately, right?

Arkatufus:
No, batched migration script will not be tested, as per our agreement

Aaronontheweb:
> No, batched migration script will not be tested, as per our agreement

Right, got it

PR #170, opened 2023-03-21
Add Csv and TagTable unit tests

Arkatufus:
* Add Sqlite and MsSqlite unit tests
* Add PostgreSql unit tests
* Add SqlServer unit tests


CumpsD:
Don't forget the copyright header for all the renamed and new files ;)

Arkatufus:
@CumpsD good catch, thanks!

Arkatufus:
The new tests are failing, I'm still looking into the cause

Arkatufus:
This PR now only contains unit tests. All the test failures are infrastructure related.
- Docker failed to run
- Journal table was missing when tests were running
- Journal table was not deleted and tests failed because of cross-talk between unit tests

Arkatufus:
Things tried:
- Move database initialization/cleanup to the start of the test
- Standardize all tests so that they uses the same initialization flow
- Move database initialization/cleanup to the end of the test
- Remove XUnit `IAsyncLifetime` and uses built-in `AfterAll` method instead
- Split the unit test for each database to use their own fixture class instead of using one monolithic fixture class

PR #175, opened 2023-03-23
Look into supporting DateTime and long databases.

### LINKED ISSUE #181: Look into supporting DateTime and long databases

### LINKED ISSUE #182: Bump Npgsql from 6.0.7 to 7.0.2

CumpsD:
Fixes https://github.com/akkadotnet/Akka.Persistence.Sql/issues/181
Fixes https://github.com/akkadotnet/Akka.Persistence.Sql/issues/182

This is a work in progress. It's an exploration to see where we would end if we had different definitions for datetime and long based providers.

The idea is to tuck away the linq2db `DataConnection` access in an `AkkaDataConnection` which is mostly just a wrapper with some extra information in which mode we are (`DateTime` vs `long`).

The exploration part is to see:
- [ ] Does it work?
- [ ] Does it allow us to cleanup some other code?
- [ ] Does it present opportunities wrt supporting different databases?

First iteration is a quick hack with exposing `.Db` to get to the raw `DataConnection`, in the upcoming commits the plan is to remove that property and expose only what is needed.

Right now it's also just dealing with the SnapshotRow, but there is equal code in the JournalRow (not the definition but the mapping), which can also benefit from this approach.

Also Fixes https://github.com/akkadotnet/Akka.Persistence.Sql/pull/171 in the mean time

### DIFF
--- a/src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
+++ b/src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
@@ -5,15 +5,21 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Data;
+using System.Threading;
 using System.Threading.Tasks;
+using Akka.Persistence.Sql.Journal.Types;
+using Akka.Persistence.Sql.Snapshot;
 using LinqToDB;
 using LinqToDB.Data;
 using LinqToDB.Data.RetryPolicy;
+using LinqToDB.SchemaProvider;
 
 namespace Akka.Persistence.Sql.Db
 {
     public class AkkaDataConnection : IDisposable, IAsyncDisposable
     {
+        private readonly DataConnection _connection;
         private readonly string _providerName;
 
         public AkkaDataConnection(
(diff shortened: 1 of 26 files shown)


CumpsD:
Removed `.Db` after the initial work was done. It does make `AkkaDataConnection` a nice central place to get an overview of which features are used.

I was doubting between 2 styles:

```csharp
await using var connection = _connectionFactory.GetConnection();

if (connection.UseDateTime)
{
	await connection
		.GetTable<DateTimeSnapshotRow>()
		.Where(r => r.PersistenceId == persistenceId)
		.DeleteAsync();
}
else
{
	await connection
		.GetTable<LongSnapshotRow>()
		.Where(r => r.PersistenceId == persistenceId)
		.DeleteAsync();
}
```

and

```csharp
await using var connection = _connectionFactory.GetConnection();

await connection.PerformAction(
	() => connection
		.GetTable<DateTimeSnapshotRow>()
		.Where(r => r.PersistenceId == persistenceId)
		.DeleteAsync(),
	() => connection
		.GetTable<LongSnapshotRow>()
		.Where(r => r.PersistenceId == persistenceId)
		.DeleteAsync());
```

@Aaronontheweb liked the first one, so went with that style in `ByteArraySnapshotDao`

CumpsD:
All tests pass with linq2db 3.5.0, will now try upgrading to 3.7.0 to see if the issue is solved

image

Aaronontheweb:
@CumpsD is this ready to be taken out of draft state?

CumpsD:
It works on linq2db 3.7.0!

image

After this is merged, I'll update linq2db to 5.1.0

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        public AkkaDataConnection(
+            string providerName,
+            DataConnection connection)
+        {
+            _providerName = providerName.ToLower();
+            _connection = connection;
+        }
+
+        public bool UseDateTime =>
```
Arkatufus: We should shortcut this to `public bool UseDateTime { get; }` and assign the bool value inside the constructor.

Arkatufus (approved):
Looks good so far, I only have a small nitpick.
It looks like we would need to do the same thing with the journal table? ie. split the `JournalRow` class into `DateTimeJournalRow` and `LongJournalRow`?

Arkatufus (approved):
LGTM

CumpsD:
Going to wait for #170 to continue working on this, since that fixes some CSV things which possibly will fix the breaking tests in this PR 🤞🏻

PR #189, opened 2023-03-29
Improve journal startup code

Arkatufus:
## Changes

- Change database initialization code to async
- Make journal startup non-blocking
- Add shutdown cancellation token to queries


src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
```diff
@@ -36,6 +37,8 @@ public class AkkaDataConnection : IDisposable, IAsyncDisposable
 
         public bool UseDateTime { get; }
 
+        public IDataProvider DataProvider => _connection.DataProvider;
```
Aaronontheweb, line 40: Need this for testing purposes only?

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -72,6 +73,7 @@ public abstract class BaseByteArrayJournalDao : BaseJournalDaoWithReadMessages,
         private static readonly Expression<Func<PersistenceIdAndSequenceNumber, long>> SequenceNumberSelector =
             r => r.SequenceNumber;
 
+        protected readonly CancellationToken ShutdownToken;
```
Aaronontheweb, line 76: Is this here to cancel any in-flight queries during the shutdown of the `ActorSystem`, essentially?

src/Akka.Persistence.Sql/Journal/Dao/ByteArrayJournalDao.cs
```diff
@@ -33,45 +37,50 @@ public sealed class ByteArrayJournalDao : BaseByteArrayJournalDao
...
-            }
-
-            if (JournalConfig.DaoConfig.SqlCommonCompatibilityMode)
+            // MS Sqlite does not support schema, we have to blindly try and create the tables
+            if (connection.DataProvider.Name is ProviderName.SQLiteMS)
             {
                 try
                 {
-                    connection.CreateTable<JournalMetaData>();
+                    await connection.CreateTableAsync<JournalRow>(token);
```
Aaronontheweb, line 52: LGTM

src/Akka.Persistence.Sql/Journal/SqlWriteJournal.cs
```diff
@@ -38,70 +38,96 @@ public static DateTime FromUnixEpochMillis(in long unixEpochMillis)
...
+
+            var config = journalConfig.WithFallback(SqlPersistence.DefaultJournalConfiguration);
+            _journalConfig = new JournalConfig(config);
+        }
+
+        protected override void PreStart()
+        {
+            base.PreStart();
+            Initialize().PipeTo(Self);
+            BecomeStacked(Initializing);
```
Aaronontheweb, line 77: Need to add a comment about why we need to use `BecomeStacked` here and not `Become`

Aaronontheweb (changes requested):
Have some minor changes

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
Arkatufus: No, sometimes we need to know what SQL provider we're working with since Linq2Db is not a magic bullet.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
Arkatufus: Yes, this is to support in-flight query cancellation.

src/Akka.Persistence.Sql/Journal/SqlWriteJournal.cs
Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #193, opened 2023-03-31
Fix query unit test, wait until write journal is ready

Arkatufus:
## Changes
Wait for the write journal to be ready before starting unit test

### DIFF
--- a/src/Akka.Persistence.Sql.Tests.Common/Query/BaseAllEventsSpec.cs
+++ b/src/Akka.Persistence.Sql.Tests.Common/Query/BaseAllEventsSpec.cs
@@ -5,24 +5,40 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Threading.Tasks;
+using Akka.Actor;
 using Akka.Configuration;
 using Akka.Persistence.Query;
 using Akka.Persistence.Sql.Config;
 using Akka.Persistence.Sql.Query;
 using Akka.Persistence.Sql.Tests.Common.Containers;
 using Akka.Persistence.TCK.Query;
+using Akka.TestKit.Extensions;
 using FluentAssertions.Extensions;
+using Xunit;
 using Xunit.Abstractions;
 
 namespace Akka.Persistence.Sql.Tests.Common.Query
 {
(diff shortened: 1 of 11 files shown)


src/Akka.Persistence.Sql.Tests.Common/Query/BaseAllEventsSpec.cs
```diff
@@ -5,25 +5,42 @@
...
             ReadJournal = Sys.ReadJournalFor<SqlReadJournal>(SqlReadJournal.Identifier);
         }
 
+        public async Task InitializeAsync()
+        {
+            // Force start read journal
+            var journal = Persistence.Instance.Apply(Sys).JournalFor(null);
+            
+            // Wait until journal is ready
+            var _ = await journal.Ask<Initialized>(IsInitialized.Instance).ShouldCompleteWithin(3.Seconds());
```
Aaronontheweb, line 38: LGTM

src/Akka.Persistence.Sql/Journal/SqlWriteJournal.cs
```diff
@@ -137,13 +137,18 @@ private bool Initializing(object message)
...
-                _writeInProgress.Remove(wf.PersistenceId);
-
-            return true;
+            switch (message)
+            {
+                case WriteFinished wf:
+                    if (_writeInProgress.TryGetValue(wf.PersistenceId, out var latestPending) & (latestPending == wf.Future))
+                        _writeInProgress.Remove(wf.PersistenceId);
+                    return true;
+                case IsInitialized:
```
Aaronontheweb, line 150: Should add a comment here explaining why this message exists

Aaronontheweb (changes requested):
One small change

src/Akka.Persistence.Sql/Journal/SqlWriteJournal.cs
Arkatufus: Good idea

Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #196, opened 2023-04-03
Clean-up table creation, add indexing to mssql

Arkatufus:
## Changes

- Clean-up table creation, use Linq2Db "create if not exist" flag
- Add indexing to SqlServer journal table

### DIFF
--- a/src/Akka.Persistence.Sql.HelperLib/TagTableMigrator.cs
+++ b/src/Akka.Persistence.Sql.HelperLib/TagTableMigrator.cs
@@ -52,10 +52,7 @@ namespace Akka.Persistence.Sql.HelperLib
             await using var connection = _connectionFactory.GetConnection();
 
             // Create the tag table if it doesn't exist
-            var dbSchema = connection.GetSchema();
-
-            if (dbSchema.Tables.All(t => t.TableName != _journalConfig.TableConfig.TagTable.Name))
-                await connection.CreateTableAsync<JournalTagRow>();
+            await connection.CreateTableAsync<JournalTagRow>(TableOptions.CreateIfNotExists);
 
             long maxId;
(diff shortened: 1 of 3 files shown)


src/Akka.Persistence.Sql.HelperLib/TagTableMigrator.cs
```diff
@@ -52,10 +52,7 @@ public async Task Migrate(long startOffset, int batchSize, long? endOffset = nul
             await using var connection = _connectionFactory.GetConnection();
 
             // Create the tag table if it doesn't exist
-            var dbSchema = connection.GetSchema();
-
-            if (dbSchema.Tables.All(t => t.TableName != _journalConfig.TableConfig.TagTable.Name))
-                await connection.CreateTableAsync<JournalTagRow>();
+            await connection.CreateTableAsync<JournalTagRow>(TableOptions.CreateIfNotExists);
```
Aaronontheweb, line 55: TagTable name is inside the `TableOptions`, correct?

Aaronontheweb (commented):
Left a question

src/Akka.Persistence.Sql.HelperLib/TagTableMigrator.cs
Arkatufus: No, `TagTable` name is inside the table mapping, this is moving the responsibility for detecting if the table exist or not from our code to the database.

Aaronontheweb: Ah I see - ok, makes sense. Just wanted to make sure I understood the significance of this change. So we're just delegating it to Linq2Db now. Got it.

Aaronontheweb (approved):
LGTM

PR #199, opened 2023-04-05
Add Akka.Hosting extension methods

Arkatufus:


src/Akka.Persistence.Sql.Hosting/Akka.Persistence.Sql.Hosting.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>$(NetStandardLibVersion)</TargetFramework>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Akka.Persistence.Hosting" />
```
Aaronontheweb, line 9: LGTM, version range for the MSFT.EXT packages is controlled by the base Akka.Hosting library.

src/Akka.Persistence.Sql.Hosting.Tests/JournalSettingsSpec.cs
```diff
@@ -0,0 +1,159 @@
...
+using Akka.Persistence.Sql.Config;
+using FluentAssertions;
+using LanguageExt.UnitsOfMeasure;
+using Xunit;
+
+namespace Akka.Persistence.Sql.Hosting.Tests
+{
+    public class JournalSettingsSpec
+    {
+        [Fact(DisplayName = "Default options should not override default hocon config")]
```
Aaronontheweb, line 11: LGTM

src/Akka.Persistence.Sql.Hosting/JournalDatabaseOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+
+using System;
+using System.Text;
+using Akka.Hosting;
+
+namespace Akka.Persistence.Sql.Hosting
+{
+    public sealed class JournalDatabaseOptions
+    {
+        public static JournalDatabaseOptions Default => new(DatabaseMapping.Default)
```
Aaronontheweb, line 15: I like the convenience of having ready-made "defaults" for each of the major supported databases here.

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
```diff
@@ -0,0 +1,252 @@
...
+        /// </param>
+        /// <returns>
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        ///     Thrown when both <paramref name="journalOptions"/> and <paramref name="snapshotOptions"/> are null.
+        /// </exception>
+        public static AkkaConfigurationBuilder WithSqlPersistence(
+            this AkkaConfigurationBuilder builder,
+            SqlJournalOptions? journalOptions = null,
```
Aaronontheweb, line 224: Won't this configuration be 100% ineffective if a connection string isn't specified? Shouldn't we disallow that on the API?

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
```diff
@@ -0,0 +1,252 @@
...
+        ///     </para>
+        ///     <b>Default</b>: <c>true</c>
+        /// </param>
+        /// <returns>
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        ///     Thrown when both <paramref name="journalOptionConfigurator"/> and <paramref name="snapshotOptionConfigurator"/> are null.
+        /// </exception>
+        public static AkkaConfigurationBuilder WithSqlPersistence(
```
Aaronontheweb, line 171: Same feedback as the other method - should we force the user to have to specify a connection string here?

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
```diff
@@ -0,0 +1,252 @@
...
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        /// <exception cref="ArgumentOutOfRangeException">
+        ///     Thrown when <see cref="journalBuilder"/> is set and <see cref="mode"/> is set to
+        ///     <see cref="PersistenceMode.SnapshotStore"/>
+        /// </exception>
+        public static AkkaConfigurationBuilder WithSqlServerPersistence(
+            this AkkaConfigurationBuilder builder,
+            string connectionString,
+            string providerName,
```
Aaronontheweb, line 78: Is it worthwhile to make this an enum rather than a magic string?

Aaronontheweb (commented):
Left some feedback on the Akka.Hosting API surface area

Aaronontheweb:
We need to add an example of using the hosting bits as well

Aaronontheweb:
I think doing some integration tests is probably the best way to go - let's use the Akka.Hosting.TestKit to do that.

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
Arkatufus: Depends, we do not control these magic string constants, Linq2Db does.
We would need to remember to update the Enum based on changes in Linq2Db on each version update.

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
Arkatufus: Connection string checking is done inside the options to HOCON transform, the method will throw on missing connection string or provider name

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
Arkatufus: This is the "advanced" version of the API, when people need to modify all aspect of their database mapping through the option classes, we already have a convenience easy to use method on line 75 of this file.

Arkatufus: Connection string checking is done inside the options to HOCON transform, the method will throw on missing connection string or provider name

src/Akka.Persistence.Sql.Hosting/SqlJournalOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+        ///     you leave this empty for greenfield projects.
+        /// </summary>
+        public JournalDatabaseOptions? DatabaseOptions { get; set; }
+        
+        protected override Configuration.Config InternalDefaultConfig => Default;
+
+        protected override StringBuilder Build(StringBuilder sb)
+        {
+            if (string.IsNullOrWhiteSpace(ConnectionString))
+                throw new ArgumentNullException(nameof(ConnectionString), $"{nameof(ConnectionString)} can not be null or empty.");
```
Arkatufus, line 75: This is where connection string value is being checked, we make sure that the user doesn't shoot themselves on the foot here.

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
Arkatufus: The API design/pattern is the same as what we have for SqlServer and PostgreSql so that all of our plugin have a similar API surface

src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
Aaronontheweb: Ok

src/Akka.Persistence.Sql.Hosting/SqlJournalOptions.cs
```diff
@@ -0,0 +1,96 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="SqlOptions.cs" company="Akka.NET Project">
```
CumpsD:
```suggestion
//  <copyright file="SqlJournalOptions.cs" company="Akka.NET Project">
```

src/Akka.Persistence.Sql.Hosting/SqlSnapshotOptions.cs
```diff
@@ -0,0 +1,82 @@
...
+// -----------------------------------------------------------------------
+
+using System;
+using System.Text;
+using Akka.Hosting;
+using Akka.Persistence.Hosting;
+
+namespace Akka.Persistence.Sql.Hosting
+{
+    public sealed class SqlSnapshotOptions: SnapshotOptions
```
CumpsD:
```suggestion
    public sealed class SqlSnapshotOptions : SnapshotOptions
```

src/Akka.Persistence.Sql.Hosting/SqlJournalOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+// -----------------------------------------------------------------------
+
+using System;
+using System.Text;
+using Akka.Hosting;
+using Akka.Persistence.Hosting;
+
+namespace Akka.Persistence.Sql.Hosting
+{
+    public sealed class SqlJournalOptions: JournalOptions
```
CumpsD:
```suggestion
    public sealed class SqlJournalOptions : JournalOptions
```

Aaronontheweb (approved):
LGTM

PR #205, opened 2023-04-10
Make snapshot table initialization asynchronous

Arkatufus:
Make snapshot table initialization more robust 

## Changes

* Make snapshot table initialization async
* Change table creation to "if not exist" instead of bruteforcing it
* Stop actor if initialization failed

### DIFF
--- a/src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
+++ b/src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
@@ -320,30 +320,16 @@ namespace Akka.Persistence.Sql.Snapshot
             }
         }
 
-        // TODO: This should be converted to async
-        public void InitializeTables()
+        public async Task InitializeTables()
         {
-            using var connection = _connectionFactory.GetConnection();
-
-            try
+            await using var connection = _connectionFactory.GetConnection();
+            if (connection.UseDateTime)
             {
-                if (connection.UseDateTime)
-                {
-                    connection.CreateTable<DateTimeSnapshotRow>();
-                }
-                else
-                {
-                    connection.CreateTable<LongSnapshotRow>();
-                }
+                await connection.CreateTableAsync<DateTimeSnapshotRow>(TableOptions.CreateIfNotExists);
             }
-            catch (Exception e)
+            else
             {
-                if (_snapshotConfig.WarnOnAutoInitializeFail)
-                {
-                    _logger.Warning(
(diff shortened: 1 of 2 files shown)


src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -320,30 +320,16 @@ await connection
...
             {
-                if (connection.UseDateTime)
-                {
-                    connection.CreateTable<DateTimeSnapshotRow>();
-                }
-                else
-                {
-                    connection.CreateTable<LongSnapshotRow>();
-                }
+                await connection.CreateTableAsync<DateTimeSnapshotRow>(TableOptions.CreateIfNotExists);
```
Arkatufus, line 328: Tell Linq2Db to generate a `IF NOT EXIST` script for the `CREATE TABLE` SQL script

src/Akka.Persistence.Sql/Snapshot/SqlSnapshotStore.cs
```diff
@@ -16,40 +16,73 @@
...
                 logger: Context.GetLogger());
+        }
+
+        public IStash Stash { get; set; }
+
+        protected override void PreStart()
+        {
+            base.PreStart();
+            Initialize().PipeTo(Self);
+            BecomeStacked(WaitingForInitialization);
```
Arkatufus, line 50: We're using `BecomeStacked` here because the receive method is sealed in parent class

src/Akka.Persistence.Sql/Snapshot/SqlSnapshotStore.cs
```diff
@@ -16,40 +16,73 @@
...
+        {
+            switch(message)
+            {
+                case Status.Success :
+                    UnbecomeStacked();
+                    Stash.UnstashAll();
+                    return true;
+                case Status.Failure msg:
+                    _log.Error(msg.Cause, "Error during {0} initialization", Self);
+                    Context.Stop(Self);
```
Arkatufus, line 63: If we fail to initialize the table, just stop the actor.

src/Akka.Persistence.Sql/Snapshot/SqlSnapshotStore.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM

PR #206, opened 2023-04-10
Fix bug in datetime column detection

Arkatufus:
## Changes

Fix missing `ToLowerInvariant()` in provider detection code


Aaronontheweb (approved):
LGTM

PR #207, opened 2023-04-10
Fix transaction isolation level to a configurable setting

Arkatufus:
## Changes

* Change journal and snapshot read/write transaction isolation mode to a configurable value to allow fine grain configuration

> NOTE:
>
> It looks like there's a bug with Linq2Db and MySql. MySql transaction could not take an unspecified isolation level, it would throw an SQL error if we do so.


Aaronontheweb (approved):
LGTM

PR #209, opened 2023-04-11
Add missing settings in Hosting extension method

Arkatufus:
## Changes

* Add missing option to set tag storage mode
* Add missing connection string and provider name settings from query


Aaronontheweb (approved):
LGTM

PR #214, opened 2023-04-13
Update Akka.Persistence.Sql.HelperLib.csproj

Arkatufus:
Make HelperLib project non-packable, it was released as a NuGet package on previous release.


Aaronontheweb (approved):
LGTM

PR #216, opened 2023-04-13
Fix snapshot table primary key and indexes

Arkatufus:
## Changes

* Add (persistence_id, sequence_nr) primary key constraint
* Add persistence_id index
* Add sequence_nr index


Aaronontheweb (approved):
LGTM

PR #221, opened 2023-04-17
Add official MySql support

Arkatufus:
## Changes
* Fix long/datetime snapshot row detection code
* Add MySql specific blob data type
* Add MySql unit tests


Aaronontheweb (approved):
LGTM

PR #222, opened 2023-04-17
Optimize SequenceNumber and Tag queries

Arkatufus:
## Changes

* Change highest sequence number query to use `.MaxAsync()` to leverage indexed column
* Change tag query to retrieve only the ordering and tag column

### DIFF
--- a/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
@@ -23,6 +23,7 @@ using Akka.Util;
 using LinqToDB;
 using LinqToDB.Data;
 using LinqToDB.Tools;
+using TaskExtensions = LanguageExt.TaskExtensions;
 
 namespace Akka.Persistence.Sql.Query.Dao
 {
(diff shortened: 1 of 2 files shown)


src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -197,11 +198,10 @@ public async Task<long> MaxJournalSequenceAsync()
             await using var connection = ConnectionFactory.GetConnection();
 
             // persistence-jdbc does not filter deleted here.
-            return await connection
+            var result = await connection
                 .GetTable<JournalRow>()
-                .Select(r => r.Ordering)
-                .OrderByDescending(r => r)
-                .FirstOrDefaultAsync();
+                .MaxAsync<JournalRow, long?>(r => r.Ordering);
```
Arkatufus, line 203: Change from
```
SELECT r.Ordering
FROM journal AS r
WHERE r.PersistenceId = @PersistenceId
ORDERBY r.Ordering DESC
TAKE 1
```

to

```
SELECT MAX(r.Ordering)
FROM journal AS r
WHERE r.PersistenceId = @PersistenceId
```

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -254,6 +254,11 @@ private static async Task AddTagDataFromTagTable(List<JournalRow> toAdd, AkkaDat
             var tagRows = await connection
                 .GetTable<JournalTagRow>()
                 .Where(r => r.OrderingId.In(toAdd.Select(row => row.Ordering).Distinct()))
+                .Select(r => new TagRow
+                {
+                    OrderingId = r.OrderingId,
+                    TagValue = r.TagValue
+                })
```
Arkatufus, line 261: Truncate query result set from 4 columns to 2

Aaronontheweb (approved):
LGTM

PR #243, opened 2023-04-27
refactor: cleanup naming, apply resharper hints

CumpsD:
Made some short variable names more descriptive, applied the `is not null` resharper advice, as well as some smaller refactorings.


Aaronontheweb (approved):
LGTM

PR #269, opened 2023-06-14
fix: add default SqlReadJournal static method

cristicazan:
## Changes

In the PersistenceQuery, in the GetDefaultConfig the framework relies on the fact that in the SqlReadJournal there is a public static method called "[GetDefaultConfig](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence.Query/PersistenceQuery.cs)". However, this is not the case int this implementation of the class and the ReadJournalFor will fail without that.


Arkatufus:
This is not quite the solution for the problem.

PR #278, opened 2023-07-14
Fix missing SQL query configuration file

Arkatufus:
Bug reported in Discord, Hosting extension does not automatically add Query default configurations

## Changes

* Append Query default HOCON configuration when journal is being used.


Aaronontheweb (approved):
LGTM

PR #283, opened 2023-07-20
Fix Persistence.Query Hosting extension support.

Arkatufus:
## Changes

Fix erroneous generated Persistence.Query hocon by `Akka.Persistence.Sql.Hosting`

### DIFF
--- a/src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
+++ b/src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
@@ -15,7 +15,7 @@ namespace Akka.Persistence.Sql.Hosting
     public static class HostingExtensions
     {
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Akka.Persistence.Sql.Hosting/HostingExtensions.cs
```diff
@@ -302,7 +302,7 @@ public static class HostingExtensions
                     builder
                         .AddHocon(journalOptions.ToConfig(), HoconAddMode.Prepend)
                         .AddHocon(journalOptions.DefaultConfig, HoconAddMode.Append)
-                        .AddHocon(SqlPersistence.DefaultConfiguration, HoconAddMode.Append),
+                        .AddHocon(journalOptions.DefaultQueryConfig, HoconAddMode.Append),
```
Aaronontheweb, line 305: Nice catch

Aaronontheweb (approved):
LGTM

PR #290, opened 2023-07-27
Make `_journalSequenceActor` instantiation `Lazy` to prevent start-up racy condition

Arkatufus:
Multiple `System.ReadJournalFor<SqlReadJournal>()` in rapid succession can throw an "actor name not unique" exception, need to defer actor creation by making it lazy.

### DIFF
--- a/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
+++ b/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
@@ -42,7 +42,7 @@ namespace Akka.Persistence.Sql.Query
 
         private readonly Source<long, ICancelable> _delaySource;
         private readonly EventAdapters _eventAdapters;
-        private readonly IActorRef _journalSequenceActor;
+        private readonly Lazy<IActorRef> _journalSequenceActor;
         private readonly ActorMaterializer _mat;
         private readonly ReadJournalConfig _readJournalConfig;
         private readonly ByteArrayReadJournalDao _readJournalDao;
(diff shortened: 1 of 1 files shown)


src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -83,12 +83,13 @@ public class SqlReadJournal :
                 // TODO: figure out a way to signal shutdown to the query executor here
                 default);
 
-            _journalSequenceActor = system.ActorOf(
-                props: Props.Create(
-                    () => new JournalSequenceActor(
-                        _readJournalDao,
-                        _readJournalConfig.JournalSequenceRetrievalConfiguration)),
-                name: $"{_readJournalConfig.TableConfig.EventJournalTable.Name}akka-persistence-sql-sequence-actor");
+            _journalSequenceActor = new Lazy<IActorRef>(
```
Aaronontheweb, line 86: If multiple journals all start up at once, won't this still be an issue, since each journal creates their own instance of the same name?

Aaronontheweb (commented):
Left a question

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
to11mtm: It looks like actually we are using `$"{configPath}.akka-persistence-sql-journal-sequence-actor"` as the actor name, so I would _think_ that it would be safer here...

_however_, this is potentially more inefficient; I would normally expect that one `JournalSequenceActor` for the table -could- be reused between different journal instances, since AFAIR it's sole purpose is to track the max sequence number...

Arkatufus: ~~Well, its lazy, it only get instantiated when the read journal needs to resequence something~~

Arkatufus: I think you're right, I'll look into it

Arkatufus: The problem starts when you tried to get multiple read journal to the same table at the same time

Arkatufus: We're caching the read journals by id inside the query plugin using a concurrent dictionary, but 2 plugins were somehow being instantiated at the same time, causing an actor name collision in the process.

The collision happened before the actor were actually created, ie. it happened when the actor name were still being reserved.

Arkatufus:
The issue stemmed from Akka.Persistence.Query, closing this PR.

PR #317, opened 2023-09-14
Fix missing default query configuration

Arkatufus:
When using HOCON configuration, default query configuration were not injected properly

## Changes

* Inject default query configuration into SqlReadJournalProvider
* Standardize SqlReadJournalProvider
* Add sample project

### DIFF
--- a/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
+++ b/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
@@ -50,8 +50,7 @@ namespace Akka.Persistence.Sql.Query
 
         public SqlReadJournal(
             ExtendedActorSystem system,
-            Configuration.Config config,
-            string configPath)
+            Configuration.Config config)
         {
             var writePluginId = config.GetString("write-plugin");
 
@@ -68,7 +67,7 @@ namespace Akka.Persistence.Sql.Query
             _mat = Materializer.CreateSystemMaterializer(
                 context: system,
                 settings: ActorMaterializerSettings.Create(system),
-                namePrefix: $"l2db-query-mat{configPath}");
+                namePrefix: $"l2db-query-mat-{Guid.NewGuid():N}");
 
             _readJournalDao = new ByteArrayReadJournalDao(
                 scheduler: system.Scheduler.Advanced,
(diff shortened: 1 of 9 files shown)


src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -68,7 +67,7 @@ public class SqlReadJournal :
             _mat = Materializer.CreateSystemMaterializer(
                 context: system,
                 settings: ActorMaterializerSettings.Create(system),
-                namePrefix: $"l2db-query-mat{configPath}");
+                namePrefix: $"l2db-query-mat-{Guid.NewGuid():N}");
```
Aaronontheweb, line 70: What was the fix here?

Aaronontheweb (commented):
Not sure I understand the source of the fix - could you help me @Arkatufus ?

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
Arkatufus: That was not the fix, the `SqlReadJournal` is a bit different in that it creates a materializer, which none of our other read journal does.

It tries to create a single materializer by appending the "config path".

The problem is that "config path" were never being passed into the provider, the provider was created through reflection magic and none of the activator code passes in the "config path".

So I just remove that ambiguous code and replace it with a GUID

src/Akka.Persistence.Sql/Query/SqlReadJournalProvider.cs
```diff
@@ -12,29 +12,17 @@ namespace Akka.Persistence.Sql.Query
...
-
-        public SqlReadJournalProvider(
-            ExtendedActorSystem system,
-            Configuration.Config config,
-            string configPath)
-        {
-            _system = system;
-            _config = config;
-            _configPath = configPath;
+            _config = config.WithFallback(SqlPersistence.DefaultQueryConfiguration);
```
Arkatufus, line 22: This is the actual fix

Aaronontheweb: Got it, ok thank you.

Aaronontheweb (approved):
LGTM

PR #324, opened 2023-09-22
Add xUnit 2.5.1 compatibility

Arkatufus:
Fixes #

## Changes

- Convert benchmark unit test methods to async
- Increase PostgreSql fixture max_connection to 1000
- Add MySql benchmark code


Aaronontheweb (approved):
LGTM

Issue #346, opened 2024-02-06
High memory usage


tudorFrizeru:
**Version Information**
AKKA 1.5.16
Modules: 
AKKA.Persistence.SQL 1.5.13, 
AKKA Streams 1.5.16, 
AKKA Serilog 1.5.12.1

**Describe the performance issue**
After migrating to AKKA 1.5, .NET 8 and the new AKKA.Persistence.SQL, the memory usage has increased to 1 GB (sometimes 2 GB between separate runs, with no changes).

Before the upgrade, the project was running AKKA 1.4.49, .NET 6 and the AKKA.Persistence.SqlServer 1.4.35 and the memory allocation was flattening at 400 MB (mainly due to asp.net core hosting).

I did the upgrade performing the complete database schema upgrade with the migration of Tags to the seaprate table and enabling the UUID in the EventJournal.

**Data and Specs**

Memory dump, sample persistence configuration and a screenshot with the diagnostic events available [here](https://1drv.ms/f/s!Ap1ENSVV1Ufai5QCqcpIzyyRJBDI0w?e=bBVSfb)

**Expected behavior**
Similar memory alocation as before 

**Actual behavior**
Memory allocation has increased to 1 GB, and doesn't increase further (as opposed to 400 MB before the upgrade), and sometimes for unexplained reasons it caps at 2 GB.

**Environment**
Running on Windows 11, .NET 8, SQL Server 2022, hosting the ActorSystem in an asp.net core environment.
Not using the new AKKA Hosting, for historical reasons I create the ActorSystem through a delegate when the asp.net server starts.
No Cluster, no Remoting.

**Additional context**
Since the migraiton, I keep seeing this line in the debug console log, every 1-2 seconds:

[22:54:19 DBG] Start timer [Akka.Persistence.Sql.Query.InternalProtocol.QueryOrderingIdsTimerKey] with generation [9]
[22:54:21 DBG] Start timer [Akka.Persistence.Sql.Query.InternalProtocol.QueryOrderingIdsTimerKey] with generation [10]
[22:54:22 DBG] Start timer [Akka.Persistence.Sql.Query.InternalProtocol.QueryOrderingIdsTimerKey] with generation [11]
[22:54:27 DBG] Start timer [Akka.Persistence.Sql.Query.InternalProtocol.QueryOrderingIdsTimerKey] with generation [12]
[...]

PR #347, opened 2024-02-10
Fix Missing take and optimize some captures

to11mtm:
May help with #346 

## Changes

- Fixes missing `TAKE` on EventsByTag Query for Tag table case.
- Provides overload for `ExecuteInTransactionAsync` that takes input state to minimize captures
- Fixed a few spots where we weren't using capture-minimizing APIs completely
- Tried to do a few other capture related optimizations and remove a double-prop access on read paths.

## Checklist

### Benchmarks

Will see what I can do...

### DIFF
--- a/src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
+++ b/src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
@@ -42,17 +42,19 @@ namespace Akka.Persistence.Sql.Db
             fmb.Build();
 
             _useCloneDataConnection = config.UseCloneConnection;
-            _opts = new DataOptions()
+            var opts =new DataOptions()
                 .UseConnectionString(config.ProviderName, config.ConnectionString)
                 .UseMappingSchema(mappingSchema);
+            _opts = opts;
 
             if (config.ProviderName.ToLower().StartsWith("sqlserver"))
                 _policy = new SqlServerRetryPolicy();
-
+            
+            
             _cloneConnection = new Lazy<AkkaDataConnection>(
                 () => new AkkaDataConnection(
-                    _opts.ConnectionOptions.ProviderName,
-                    new DataConnection(_opts)));
+                    opts.ConnectionOptions.ProviderName,
+                    new DataConnection(opts)));
         }
 
         public AkkaPersistenceDataConnectionFactory(IProviderConfig<SnapshotTableConfiguration> config)
@@ -81,17 +83,17 @@ namespace Akka.Persistence.Sql.Db
             fmb.Build();
 
             _useCloneDataConnection = config.UseCloneConnection;
-            _opts = new DataOptions()
+            var opts = new DataOptions()
                 .UseConnectionString(config.ProviderName, config.ConnectionString)
                 .UseMappingSchema(mappingSchema);
-
+            _opts = opts;
             if (config.ProviderName.ToLower().StartsWith("sqlserver"))
                 _policy = new SqlServerRetryPolicy();
 
             _cloneConnection = new Lazy<AkkaDataConnection>(
                 () => new AkkaDataConnection(
-                    _opts.ConnectionOptions.ProviderName,
-                    new DataConnection(_opts)));
+                    opts.ConnectionOptions.ProviderName,
+                    new DataConnection(opts)));
         }
 
         private static void MapJournalRow(
--- a/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
+++ b/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
@@ -9,6 +9,7 @@ using System.Data;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Persistence.Sql.Db;
+using Akka.Persistence.Sql.Query.Dao;
 
 namespace Akka.Persistence.Sql.Extensions
 {
@@ -71,6 +72,14 @@ namespace Akka.Persistence.Sql.Extensions
 
                 throw;
             }
+        }
+        
+        internal static Task<T> ExecuteWithTransactionAsync<TState,T>(
+            this DbStateHolder factory,
+            TState state,
+            Func<AkkaDataConnection, CancellationToken, TState, Task<T>> handler)
+        {
+            return factory.ConnectionFactory.ExecuteWithTransactionAsync(state, factory.IsolationLevel, factory.ShutdownToken, handler);
         }
         
         public static async Task<T> ExecuteWithTransactionAsync<TState,T>(
--- a/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
+++ b/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
@@ -72,5 +72,36 @@ namespace Akka.Persistence.Sql.Extensions
                 throw;
             }
         }
+        
+        public static async Task<T> ExecuteWithTransactionAsync<TState,T>(
+            this AkkaPersistenceDataConnectionFactory factory,
+            TState state,
+            IsolationLevel level,
+            CancellationToken token,
+            Func<AkkaDataConnection, CancellationToken, TState, Task<T>> handler)
+        {
+            await using var connection = factory.GetConnection();
+            await using var tx = await connection.BeginTransactionAsync(level, token);
+
+            try
+            {
+                var result = await handler(connection, token, state);
+                await tx.CommitAsync(token);
+                return result;
+            }
+            catch (Exception ex1)
+            {
+                try
+                {
+                    await tx.RollbackAsync(token);
+                }
+                catch (Exception ex2)
+                {
+                    throw new AggregateException("Exception thrown when rolling back database transaction", ex2, ex1);
+                }
+
+                throw;
+            }
+        }
     }
 }
--- a/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
@@ -84,7 +84,9 @@ namespace Akka.Persistence.Sql.Query.Dao
             {
                 TagMode.Csv => AsyncSource<JournalRow>
                     .FromEnumerable(
-                        new { args= new QueryArgs(offset,maxOffset,maxTake,tag,TagMode.Csv), _connectionFactory = ConnectionFactory },
+                        new { args= new QueryArgs(offset,maxOffset,maxTake,
+                            $"{separator}{tag}{separator}"
+                            ,TagMode.Csv), _connectionFactory = ConnectionFactory },
                         async input =>
                         {
                             //var tagValue = input.tag;
--- a/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
@@ -31,6 +31,7 @@ namespace Akka.Persistence.Sql.Query.Dao
         private readonly Flow<JournalRow, Try<(IPersistentRepresentation, IImmutableSet<string>, long)>, NotUsed> _deserializeFlow;
 
         private readonly ReadJournalConfig _readJournalConfig;
+        private readonly TagMode _tagMode;
 
         protected BaseByteReadArrayJournalDao(
             IAdvancedScheduler scheduler,
@@ -42,6 +43,7 @@ namespace Akka.Persistence.Sql.Query.Dao
             : base(scheduler, materializer, connectionFactory, readJournalConfig, shutdownToken)
         {
             _readJournalConfig = readJournalConfig;
+            _tagMode = _readJournalConfig.PluginConfig.TagMode;
             _deserializeFlow = serializer.DeserializeFlow();
         }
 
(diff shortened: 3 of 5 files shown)


src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -42,6 +43,7 @@ public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessag
             : base(scheduler, materializer, connectionFactory, readJournalConfig, shutdownToken)
         {
             _readJournalConfig = readJournalConfig;
+            _tagMode = _readJournalConfig.PluginConfig.TagMode;
```
to11mtm: Idea here is we don't want to repeatedly read this prop.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -239,35 +246,51 @@ private static int MaxTake(long max)
...
-                        });
+                    var cf = input._connectionFactory;
+                    var a = input.args;
+                    return await ExecuteEventQuery(a);
                 }
             ).Via(_deserializeFlow);
         }
+        
+        
+        internal async Task<List<JournalRow>> ExecuteEventQuery(QueryArgs queryArgs)
```
to11mtm: This is intentionally moved into it's own method,

Among other things, Another memory optimization for us to consider is -not- using Akka streams in the `Unfold` internal read calls; Where this gets tricky, is that marshalling via Akka streams pipeline, neatly solves problems where a lot of DB Providers aren't really all that async 🙃. 

SQLite (Fair enough,) I think some oracle providers, IDK about MySQL...

I can fold back in for now if needed, but it does make the pipeline at least a little bit neater.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -239,35 +246,51 @@ private static int MaxTake(long max)
...
-                                        r.Ordering <= input.maxOffset &&
-                                        r.Deleted == false)
-                                .OrderBy(r => r.Ordering)
-                                .Take(input.maxTake);
-
-                            return await AddTagDataIfNeededAsync(query, connection, token);
-                        });
+                    var cf = input._connectionFactory;
+                    var a = input.args;
+                    return await ExecuteEventQuery(a);
```
to11mtm: Needs Clean.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -82,53 +84,57 @@ public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessag
...
+                                                .Where(
+                                                    jtr =>
+                                                        jtr.OrderingId > txInput.Offset
+                                                        && jtr.OrderingId <= txInput.MaxOffset
+                                                        && jtr.TagValue == txInput.Tag),
+                                            SqlJoinType.Left,
+                                            (jr, jtr) => (jr.Ordering == jtr.OrderingId),
+                                            (jr, jtr) => jr)
+                                        .OrderBy(r => r.Ordering)
+                                        .Take(txInput.Max);
```
to11mtm: So, We didn't have a take here.

~~However, AFAIR, the LINQ query syntax tends to be more finicky with captures/closures than (_carefully_) plotted out L2Db Method syntax. However, if the lambda syntax somehow is producing something needed for a specific DB, we can go back to it.~~ NVM It's not playing nice and we can deal with that later.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -239,35 +246,51 @@ private static int MaxTake(long max)
...
+
+                    if (a.TagMode != TagMode.TagTable)
+                    {
+                        return await query.ToListAsync(token);        
+                    }
+                    else
+                    {
+                        return await AddTagDataFromTagTableAsync(query, connection, token);
+                    }
+                    //return await AddTagDataIfNeededAsync(query, connection, token);
```
to11mtm: clean up commented code.

src/Akka.Persistence.Sql/Query/Dao/QueryArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+using Akka.Persistence.Sql.Config;
+
+namespace Akka.Persistence.Sql.Query.Dao
+{
+    internal readonly struct QueryArgs
+    {
+        public readonly long Offset;
+        public readonly long MaxOffset;
+        public readonly int Max;
+        public readonly TagMode Mode;
```
to11mtm: IDK if I like it here, but It doesn't cost us anything on x64 AFAIK.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -146,24 +152,25 @@ orderby r.Ordering
             => Task.FromResult(
                 AsyncSource<JournalRow>
                     .FromEnumerable(
-                        new { _connectionFactory = ConnectionFactory, persistenceId, fromSequenceNr, toSequenceNr, toTake = MaxTake(max) },
+                        new {  persistenceId, fromSequenceNr, toSequenceNr, toTake = MaxTake(max) },
```
to11mtm: This whole bit is making me ask if we need a better way to capture certain 'needful' for these captures.
What I'm wondering is, should we have a helper class like:

```
public class CaptureHolder
{
     public readonly AkkaPersistenceDataConnectionFactory;
     public readonly ??? ReadIsolationLevel;
     public readonly CancellationToken ShutdownToken;
     public readonly TagMode Mode;
}
```

Then, we can pass that reference, instead of capturing `this`...
And frankly that might help further 'separate' the query logic out for the sake of testability. :)

to11mtm (commented):
Self review with some food for thought.

src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -42,17 +42,19 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<JournalTableConfig>
...
                 _policy = new SqlServerRetryPolicy();
-
+            
+            
             _cloneConnection = new Lazy<AkkaDataConnection>(
                 () => new AkkaDataConnection(
-                    _opts.ConnectionOptions.ProviderName,
-                    new DataConnection(_opts)));
+                    opts.ConnectionOptions.ProviderName,
+                    new DataConnection(opts)));
```
to11mtm, line 57: subtle thing; now we are capturing just the `DataOptions` reference rather than the factory itself...

IDK how much the GC *really* cares but it sounds like the sort of thing that can annoy a cycle detector,

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -82,53 +84,57 @@ public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessag
             {
                 TagMode.Csv => AsyncSource<JournalRow>
                     .FromEnumerable(
-                        new { separator, tag, offset, maxOffset, maxTake, _connectionFactory = ConnectionFactory },
+                        new { args= new QueryArgs(offset,maxOffset,maxTake,tag,TagMode.Csv), _connectionFactory = ConnectionFactory },
```
to11mtm: this got broken in cleanup, needs separator added.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -82,54 +84,56 @@ public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessag
...
                                 {
-                                    var journalTable = connection.GetTable<JournalRow>();
-                                    var tagTable = connection.GetTable<JournalTagRow>();
-
                                     var query =
-                                        from r in journalTable
-                                        from lp in tagTable.Where(jtr => jtr.OrderingId == r.Ordering).DefaultIfEmpty()
-                                        where lp.OrderingId > input.offset &&
-                                              lp.OrderingId <= input.maxOffset &&
+                                        from r in connection.GetTable<JournalRow>()
```
to11mtm, line 120: It would still be nice to convert this to method syntax later, if only because it's something like 300 lines of IL as it stands 😅

src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
```diff
@@ -72,5 +73,44 @@ public static class ConnectionFactoryExtensions
...
             }
         }
+        
+        internal static Task<T> ExecuteWithTransactionAsync<TState,T>(
+            this DbStateHolder factory,
+            TState state,
+            Func<AkkaDataConnection, CancellationToken, TState, Task<T>> handler)
+        {
+            return factory.ConnectionFactory.ExecuteWithTransactionAsync(state, factory.IsolationLevel, factory.ShutdownToken, handler);
+        }
```
to11mtm, line 83: Convenience method; since we are using the stateholder to avoid other captures, may as well add some sugar to go with it and make the main DAO cleaner.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -239,35 +233,59 @@ private static int MaxTake(long max)
...
-                        });
+                    return await ExecuteEventQuery(input._dbStateHolder, input._dbStateHolder.Mode, input.args);
                 }
             ).Via(_deserializeFlow);
         }
+        
+        
+        internal static async Task<List<JournalRow>> ExecuteEventQuery(DbStateHolder stateHolder, TagMode tagMode, QueryArgs queryArgs)
+        {
+            return tagMode != TagMode.TagTable
```
to11mtm, line 247: This is ugly and IDK how I feel but it's the easiest way to get more captures out of the way, short of some more significant refactoring where we had some different ways to compose without going back to `Func<>` madness...

src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
Aaronontheweb: LGTM

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -50,20 +52,19 @@ public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessag
...
+                new { _dbStateHolder, maxTake },
+                static async input =>
                 {
-                    return await input._connectionFactory.ExecuteWithTransactionAsync(
-                        ReadIsolationLevel,
-                        ShutdownToken,
-                        async (connection, token) =>
+                    return await input._dbStateHolder.ExecuteWithTransactionAsync(
+                        input.maxTake,
+                        async (connection, token,take) =>
```
Aaronontheweb, line 60: Can we just return a `Task` here instead of `async` / `await` ?

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
Aaronontheweb: > This is ugly 

Use an `if` statement instead of a ternary operator and it'll look better, I swear :p

src/Akka.Persistence.Sql/Query/Dao/DbStateHolder.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Threading;
+using Akka.Persistence.Sql.Config;
+using Akka.Persistence.Sql.Db;
+
+namespace Akka.Persistence.Sql.Query.Dao
+{
+    /// <summary>
+    /// Used to help improve capture usage and ease composition via extension methods.
+    /// </summary>
+    internal class DbStateHolder
```
Aaronontheweb: mark as `sealed` for even more IL improvements

Aaronontheweb (approved):
Generally LGTM, but did leave some comments

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
to11mtm: I'm honestly not sure, concerns aside from potential stack trace confusion on errors would be whether the fact this func is actually invoked inside two `using` contexts... Better safe than sorry when it comes to that stuff IMO.

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
to11mtm: The comment that was somehow left unposted, I wound up moving these back into separate methods.

src/Akka.Persistence.Sql/Query/Dao/DbStateHolder.cs
to11mtm: dunzo

to11mtm:
@Aaronontheweb  if you and @Arkatufus are OK with this feel free to merge, otherwise LMK what you would like to see changed. <3

src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
Arkatufus: Some discussions on this are here: 
- https://stackoverflow.com/questions/24110213/difference-between-returning-and-awaiting-a-task-in-an-async-method
- https://code-maze.com/charp-difference-between-returning-and-awaiting-a-task/

Would the performance boost for each query execution be worth it, does it outweight the possible exception stack trace confusion?

From my experience, the "proper" stack trace from an exception thrown from within an awaited Task is confusing in itself, since it is thrown from inside the async...await FSM and most often it will be a detached from the actual code that invoked it (when the thrown exception is caught by the debugger), so I can't really say that it is an actual improvement in debugging experience. But then again, I could be doing it wrong.

On the other hand, the article does point out that using proper async puts the exception where it belongs, in the future when it is supposed to happen, if it indeed needs to be thrown.

Arkatufus: I tend to agree with @to11mtm on this one, returning a `Task` from inside a `using` block is quite dangerous because we don't know just how robust the `IDispose` is being implemented by the 3rd party API developers.

Arkatufus (approved):
LGTM

PR #415, opened 2024-07-08
Fixes bug with long delay with initialization of JournalSequenceActor

### LINKED ISSUE #345: CurrentEventsByTag query on Sqlite has a weird initial delay
**Version Information**
I'm using Akka.Cluster.Hosting v1.5.15 and the associated libraries like Akka 1.5.15.

**Describe the performance issue**
I'm using an ASP.NET Core background service that uses a PersistenceQuery (CurrentEventsByTag) that should be run at application start to retrieve some data and cache it in-memory. I want my app to scale using PostgreSQL but also scale down using Sqlite which works great by using Akka.Persistence.Sql and making the provider itself configurable.

The background service I'm using looks like this and is added to the ASP.NET Core app via AddHostedService.

```csharp
public class TestBackgroundService : BackgroundService
{
    private readonly ActorSystem _actorSystem;

    public TestBackgroundService(ActorSystem actorSystem)
    {
        _actorSystem = actorSystem;
    }

    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        Log.Information("TestBackgroundService: Starting query...");

        var query = PersistenceQuery
            .Get(_actorSystem)
            .ReadJournalFor<SqlReadJournal>(SqlReadJournal.Identifier)
            .CurrentEventsByTag("t:wfd", Offset.NoOffset());

        await foreach (var ee in query.RunAsAsyncEnumerable(_actorSystem).WithCancellation(stoppingToken))
        {
            Log.Information("TestBackgoundService: Got event {Event}", ee.Event);
        }
    }
}
```

When using the Sqlite provider (with Microsoft.Data.Sqlite v8.0.1), I get my first event after 11 seconds:

```
[17:25:24 INF] TestBackgroundService: Starting query...
[17:25:24 INF] Now listening on: http://localhost:12341
[17:25:24 INF] Application started. Press Ctrl+C to shut down.
[17:25:24 INF] Hosting environment: Development
[17:25:24 INF] Content root path: /Users/kp/Projects/test/src/test
[17:25:35 INF] TestBackgoundService: Got event Incremented { Tag = t:wfd }
[17:25:35 INF] TestBackgoundService: Got event Incremented { Tag = t:wfd }
```

This delay doesn't happen with PostgreSQL or using another query type. Is there a reason the CurrentEventsByTag query runs after a 10-11s delay? The Sqlite database itself it empty except for the two events in the journal. I'm using the tag table instead of csv tags. 

Could you give me a tip on how to get more information of the persistence query performance?

Also I expect to be the Sqlite provider to be slower than the PostgreSQL provider and a 10s delay is not the end of the world for an ASP.NET Core web app - but this doesn't feel like just "bad performance" but more like theres an intentional 10s delay somewhere and I would like to understand why.

**Environment**

I'm using a 16-inch 2021 MacBook Pro (M1) with 16 GB RAM.
.NET SDK:
 Version:           8.0.101
 Commit:            6eceda187b
 Workload version:  8.0.100-manifests.8a1da915

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  14.0
 OS Platform: Darwin
 RID:         osx-arm64
 Base Path:   /usr/local/share/dotnet/sdk/8.0.101/

lucavice:
Fixes #345

## Changes

By studying the `SqlReadJournal` and `JournalSequenceActor` initialization code, I found a bug that caused a longer than expected delay for the reader to start replaying current events in the journal. The initial intended logic seemed correct, which is:

1. When the `JournalSequenceActor` is booted up, it starts by detecting gaps in the ordering numbers in the journal in batches, starting from the very beginning. If a gap is found, it will wait for up to `QueryDelay*MaxTries` for the gap to be filled, in case another process is writing into the journal.
2. In order to avoid replaying everything from the very beginning, a message `AssumeMaxOrderingId` is scheduled at the same time with a delay of `QueryDelay*MaxTries`, which is the same as the maximum wait for gaps. When this message arrives, if the gaps being analyzed have not caught up yet to the total max of the journal, it will force to update the current ordering being inspected to be the max available in the journal. This makes perfect sense, as it will guarantee that when the `SqlReadJournal` boots up, it will wait at max `QueryDelay*MaxTries` for any gaps to be filled, regardless of the length of the journal.

Unfortunately, due to a bug using an incorrect parameter, step 2 was not working as intended. This means that for journal with large number of events and gaps, the initialization could take a very long time.

The change only fixes the supplied parameter with the original intent, which is to force the scanned ordering number to be the max after the configured delay. It was instead using `_maxTries`, which is simply the number of tries to wait for gap (by default this value is 10).

In my tests, after this change the maximum amount of wait at that I experience at `SqlReadJournal` initialization is always the expected 10s (which is QueryDelay [1s] * MaxTries [10]). Previously, it could take several minutes or even hours.

### DIFF
--- a/src/Akka.Persistence.Sql/Query/JournalSequenceActor.cs
+++ b/src/Akka.Persistence.Sql/Query/JournalSequenceActor.cs
@@ -71,7 +71,7 @@ namespace Akka.Persistence.Sql.Query
                             o =>
                                 ReceiveHandler(
                                     o,
-                                    _maxTries,
+                                    a.Max,
                                     missingByCounter,
                                     moduloCounter,
                                     previousDelay));


src/Akka.Persistence.Sql/Query/JournalSequenceActor.cs
```diff
@@ -71,7 +71,7 @@ private bool ReceiveHandler(object message)
                             o =>
                                 ReceiveHandler(
                                     o,
-                                    _maxTries,
+                                    a.Max,
```
to11mtm, line 74: 🤦 

Nice catch. 😅

to11mtm (approved):
Thank you so much for digging into this!

Looks good and good write-up. <3

Aaronontheweb (approved):
LGTM

lucavice:
Thanks everybody for merging quickly and releasing 1.5.26 with the fix. I have deployed it to our staging environment and I can confirm that it now works fine with just the expected small initial delay. 

I have also been running load testing on our environment to compare performance between `Akka.Persistence.Sql` and the older `Akka.Persistence.SqlServer`. We have seen some fairly dramatic performance improvements on our use case, even without using the Tag Table. If it's of any interest to you, I could share with you some charts, but I'm not sure where (it does not fit the topics of issues?).

Aaronontheweb:
>  If it's of any interest to you, I could share with you some charts, but I'm not sure where (it does not fit the topics of issues?).

@lucavice yes, that would be wonderful if you could share that

PR #426, opened 2024-07-30
Add EventEnvelope Tags support

Arkatufus:
## Changes

* Enable EventEnvelope tags support in query result
* Enable tags checking in Persistence.TCK unit tests

### DIFF
--- a/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
@@ -170,8 +170,8 @@ namespace Akka.Persistence.Sql.Query.Dao
                         {
                             try
                             {
-                                var (representation, _, ordering) = t.Get();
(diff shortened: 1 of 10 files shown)


src/Akka.Persistence.Sql/Query/Dao/BaseByteReadArrayJournalDao.cs
```diff
@@ -28,7 +28,7 @@ namespace Akka.Persistence.Sql.Query.Dao
 {
     public abstract class BaseByteReadArrayJournalDao : BaseJournalDaoWithReadMessages, IReadJournalDao
     {
-        private readonly Flow<JournalRow, Try<(IPersistentRepresentation, IImmutableSet<string>, long)>, NotUsed> _deserializeFlow;
```
Aaronontheweb, line 31: Good call on moving from immutable sets to arrays - this will eliminate any additional allocations when we're trying to pass tag collections into the `EventEnvelope`.

Aaronontheweb (approved):
LGTM

PR #434, opened 2024-08-11
WIP Snapshot Load Batch Queue

### LINKED ISSUE #432: Bug: can't support multiple snapshots taken at same `seqNr` and `persistentId`
**Version Information**
Version of Akka.NET? v1.5.27 and later

**Describe the bug**

Related issue: https://github.com/akkadotnet/akka.net/issues/7312 - the problem is the primary key index we use for the `SnapshotStore` doesn't also cover the `Timestamp`, thus we can end up with the issues described in that thread.

**Expected behavior**

Should be able to save snapshots at the same `PersisentId` and `SeqNr` combination because they will typically be differentiated by `Timestamp` value.

**Actual behavior**

When `SaveSnapshot` call volume is low, things will still work because the snapshot code is designed to replace the existing snapshot at the matching `seqNr` through simulated "upsert" mechanics. However, this is vulnerable to race conditions if multiple snapshots in a row are all saved in rapid succession.

**Environment**

SQL Server

**Additional context**

I added work-around ideas to https://github.com/akkadotnet/akka.net/issues/7312 - that issue must be fixed first before this one can be closed out.

to11mtm:
Will not fix #432 but will probably help on some level

## Changes

This PR changes Getting default snapshot store load operation of getting the 'newest' snapshot (i.e. not caring about SEQ or timestamp) from being a multitude of snapshot requests into a queue, similar to how we handle Write requests on Journals.

## Checklist

### Latest `dev` Benchmarks 

We don't really have benchmarks against snapshot. if someone wants to add I'm happy to rebase/etc and test

### This PR's Benchmarks

See above.

### Stuff I still need to validate:

The main difference of course is that this is a -read- and more complex than a sequence number [0], and as such I did take the -small- step of splitting out deserialization from the main read logic as a separate stage.

This may have a minor performance impact on cases where only a small number of actors are recovering, however it should greatly improve recovery performance under load. Overall intent is to have the numbers be configurable for a given case (maybe we have a fallback switch for those who want old behavior?)

[0] - Hint hint ;)

### DIFF
--- a/src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
+++ b/src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
@@ -5,20 +5,49 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Linq;
 using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 using Akka.Event;
 using Akka.Persistence.Sql.Config;
 using Akka.Persistence.Sql.Db;
 using Akka.Persistence.Sql.Extensions;
 using Akka.Streams;
+using Akka.Streams.Dsl;
 using Akka.Util;
 using LinqToDB;
+using LinqToDB.Tools;
 
 namespace Akka.Persistence.Sql.Snapshot
 {
+    public class LatestSnapRequestEntry
+    {
+        public LatestSnapRequestEntry(string persistenceId)
+        {
+            PersistenceId = persistenceId;
+            TCS = new TaskCompletionSource<Option<SelectedSnapshot>>(TaskCreationOptions.RunContinuationsAsynchronously);
+        }
+
+        public readonly string PersistenceId;
+        public readonly TaskCompletionSource<Option<SelectedSnapshot>> TCS;
+    }
+
+    public class QueryLatestSnapSet
+    {
+        public readonly Dictionary<string, List<TaskCompletionSource<Option<SelectedSnapshot>>>> Entries = new();
+
+        public void Add(LatestSnapRequestEntry entry)
+        {
+            if (Entries.TryGetValue(entry.PersistenceId, out var item) == false)
+            {
+                item = Entries[entry.PersistenceId] = new List<TaskCompletionSource<Option<SelectedSnapshot>>>();
+            }
+            item.Add(entry.TCS);
+        }
+    }
     public class ByteArraySnapshotDao : ISnapshotDao
     {
         private readonly AkkaPersistenceDataConnectionFactory _connectionFactory;
@@ -29,6 +58,8 @@ namespace Akka.Persistence.Sql.Snapshot
         private readonly CancellationTokenSource _shutdownCts;
         private readonly SnapshotConfig _snapshotConfig;
         private readonly IsolationLevel _writeIsolationLevel;
+        private readonly Channel<LatestSnapRequestEntry> _pendingLatestChannel;
+        private readonly Task<Done> _latestSnapStream;
 
         public ByteArraySnapshotDao(
             AkkaPersistenceDataConnectionFactory connectionFactory,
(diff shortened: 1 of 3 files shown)


to11mtm:
This doesn't pipeline writes, but it does pipeline reads which...

 1. is very useful for performance in real world, alas writing a benchmark for true 'contentious' states is not fun... but similar to how write batching helps there, this -will- help read batching, it's mostly a matter of right-sizing the params for the stream for your overall case (i.e. latency vs throughput is dependent on size and number of snapshots... big thing here is we may not be faster for all cases but should be more *stable*, but still tunable to get faster than before in almost all cases.

tl;dr- this is a change on reads will help overall perf in a real-world system most likely.

 2. Is an example of how one -could- take the example further for other reads (e.x. expand this concept for other snapshot queries or perhaps for journal sequences, I would say 'other reads' but general journal reads are tricky for lots of reasons...)

tl;dr- we can do this pattern or a variant in other places for reads, and it will help with performance under load scenarios.

 3. Is admittedly not optimized for all SQL cases. i.e. depending on dialect there are better ways to 'do' this, they aren't that ugly/hard to wire up. I didn't do them because I am trying to get better about overdoing first revisions. >_< But I know PG for instance can do -way- better using `DISTINCT ON`, see https://stackoverflow.com/a/34715134 . Also the 'what do I need to `Option.None`' logic could use improvement.

tl;dr- this PR is a basic implementation and may require some optimization to do the 'best thing' for each database. Or in general

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -5,20 +5,77 @@
...
 namespace Akka.Persistence.Sql.Snapshot
 {
+    public static class SubFlowExtensions
+    {
+        public static Source<TOut, TMat> MergeSubStreamsAsSource<TOut, TMat,
+            TClosed>(this SubFlow<TOut, TMat, TClosed> subFlow)
+        {
+            return (Source<TOut,TMat>)(subFlow.MergeSubstreams());
+        }
+    }
```
to11mtm, line 33: Happy for feedback on this method but TBH it would be nice if some form (even if it's not returning `Source<TOut,TMat>`) was put into akka streams proper. It's _really_ useful

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -5,20 +5,77 @@
...
+        public QueryLatestSnapSet a { get; }
+        public List<LongSnapshotRow> b { get; }
+        public Exception? err { get; }
+        public void Deconstruct(out QueryLatestSnapSet a, out List<LongSnapshotRow> b, out Exception? err)
+        {
+            a = this.a;
+            b = this.b;
+            err = this.err;
+        }
+    }
```
to11mtm, line 66: needs cleanup also not sure if struct is actually good idea or not here :)

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -5,20 +5,77 @@
...
+
+        public void Add(LatestSnapRequestEntry entry)
+        {
+            if (Entries.TryGetValue(entry.PersistenceId, out var item) == false)
+            {
+                item = Entries[entry.PersistenceId] = new List<TaskCompletionSource<Option<SelectedSnapshot>>>();
+            }
+            item.Add(entry.TCS);
+        }
+    }
```
to11mtm: May or may not be better to return `this` on `.Add()` for cosmetics/etc.

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -48,6 +107,181 @@ public class ByteArraySnapshotDao : ISnapshotDao
             _readIsolationLevel = snapshotConfig.ReadIsolationLevel;
 
             _shutdownCts = new CancellationTokenSource();
+            _pendingLatestChannel = Channel.CreateUnbounded<LatestSnapRequestEntry>();
+            int maxSubStreamsForReads = 8; // TODO: Configurable
+            int maxRequestsPerBatch = 50;
```
to11mtm: These should be configurable

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -48,6 +107,181 @@ public class ByteArraySnapshotDao : ISnapshotDao
...
+            _latestSnapStream = Source.ChannelReader(_pendingLatestChannel.Reader)
+                .GroupBy(maxSubStreamsForReads, a=> a.PersistenceId.GetHashCode()% maxSubStreamsForReads)
+                .BatchWeighted(
+                    maxRequestsPerBatch,
+                    a => 1,
```
to11mtm, line 122: I know in ShotGlass I had/have a Batcher that could assign a 'zero weight' for cases where a persistence ID was already queued...

FBOW in this first go here I instead decided a flat weight of 1 would be a lot less 'change' but also has -some- benefits for forward progress. (but really mostly because a custom stream stage felt like overkill for first intro of this.)

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -48,6 +107,181 @@ public class ByteArraySnapshotDao : ISnapshotDao
...
+                    })
+                .SelectAsync(1,
+                    async a =>
+                    {
+
+                        using (var connection = _connectionFactory.GetConnection())
+                        {
+                            if (connection.UseDateTime)
+                            {
+                                //TODO: Consolidate/fixup different rowtype issues.
```
to11mtm, line 142: I am admittedly unsure how we got to having separate categories here; we -should- be able to set all of this up such that either L2DB is doing the 'right thing' for a given db config, and/or we sanely do our queries against whatever mapping is there. (well, actually it's a bit of both... still...)

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -48,6 +107,181 @@ public class ByteArraySnapshotDao : ISnapshotDao
...
+                                            }).ToListAsync();
+                                    return new (a, set, err: (Exception?)null);
+                                }
+                                catch (Exception ex)
+                                {
+                                    return new (a, null, err: ex);
+                                }
+                            }
+                        }
+                    }).SelectAsync(1,
```
to11mtm, line 212: IDK if this should -really- be `SelectAsync`, alas I don't have a clean way to do `.Async()` on a Subflow ;_; ???

In general the _desire_ here is to ensure that we can fire off results to things without blocking the next read. Probably a cleaner/better way to do this, open to suggestions. <3

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -48,6 +107,181 @@ public class ByteArraySnapshotDao : ISnapshotDao
...
+                                {
+                                    if (a.Entries.TryGetValue(se, out var setNo))
+                                    {
+                                        foreach (var taskCompletionSource in setNo)
+                                        {
+                                            taskCompletionSource.TrySetResult(Option<SelectedSnapshot>.None);
+                                        }
+                                    }
+                                }
+                            }
```
to11mtm: I hate this and would love cleaner suggestions. <3

src/Akka.Persistence.Sql/Snapshot/ByteArraySnapshotDao.cs
```diff
@@ -182,41 +416,50 @@ await connection
                 });
         }
 
-        public async Task<Option<SelectedSnapshot>> LatestSnapshotAsync(
+        public Task<Option<SelectedSnapshot>> LatestSnapshotAsync(
             string persistenceId,
             CancellationToken cancellationToken = default)
         {
-            var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _shutdownCts.Token);
```
to11mtm, line 189: This should probably pass the CTS or whatever into the request, for the sake of allowing the token to do the needful.

to11mtm (commented):
Added my self-comments, just in case I get bussed/whatever.

PR #447, opened 2024-08-30
Modernize benchmark tests

Arkatufus:
## Changes

* Update all benchmark figures in the documentation
* Add MathNet.Numerics package to Benchmark.Tests
* Update SqlServerContainer to 2022-latest and SqlServer.2022 provider name
* Update PostgreSqlContainer to PostgreSQL.15 provider name
* Change Benchmark.Tests journal performance calculation to use outlier rejection
  * Outlier rejection is based on Z-score method
  * Mean and standard deviation is calculated
  * Minimum and maximum acceptable values are calculated as [ mean +- (sigma * StdDev) ]
* Performance table values are the median value of non-rejected data points.

### DIFF
--- a/docs/articles/benchmarks.md
+++ b/docs/articles/benchmarks.md
@@ -5,7 +5,75 @@ title: Performance Benchmarks
 
 # Performance Benchmarks
 
-## Persistence Performance
+## Journal Events Persistence Performance
+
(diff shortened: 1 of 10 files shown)


Aaronontheweb (approved):
LGTM

Aaronontheweb (approved):
LGTM

PR #448, opened 2024-08-30
Port akkadotnet/akka.net#7313

Arkatufus:
Port of akkadotnet/akka.net#7313

## Changes

Make sure that `ByteArraySnapshotDao.DeleteAsync()` respects timestamp value
Regression unit tests were not needed because it is already part of the new Akka.Persistence.TCK 1.5.28-beta1


Aaronontheweb (approved):
lgtm

PR #450, opened 2024-09-04
#432 add specs

Arkatufus:
Add unit test specs for #432

> [!NOTE]
> 
> This should probably moved to Akka.Persistence.TCK so it can be propagated to all of Akka.Persistence.* packages

## Changes

Add unit test specs, will fail until SaveSnapshot concurrency problem is solved


Arkatufus:
Will be replaced by unit tests in Akka.Persistence.TCK in v1.5.28

PR #455, opened 2024-09-05
Add SnapshotStore save snapshot spec

### LINKED ISSUE #432: Bug: can't support multiple snapshots taken at same `seqNr` and `persistentId`

Arkatufus:
Fixes #432

## Changes

Implement spec for bug (included in Persistence.TCK 1.5.28), actual fix was in #448

### DIFF
--- a/src/Akka.Persistence.Sql.Tests/MySql/MySqlSnapshotStoreSaveSnapshotSpec.cs
+++ b/src/Akka.Persistence.Sql.Tests/MySql/MySqlSnapshotStoreSaveSnapshotSpec.cs
@@ -7,6 +7,7 @@
 using System;
 using Akka.Configuration;
 using Akka.Persistence.Sql.Tests.Common.Containers;
+using Akka.Persistence.Sql.Tests.Common.Internal.Xunit;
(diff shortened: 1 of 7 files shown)


Aaronontheweb (approved):
LGTM

PR #456, opened 2024-09-05
Adds DataOptionsSetup

### LINKED ISSUE #18: Create `Setup` Classes to allow configuration without writing real HOCON
With other parts of Akka.Net having `ActorSystemSetup` and similar, it would be nice to have the same capabilities here so that users can easily set up plugin configurations without having to write a bunch of HOCON.

This class should allow for configuration of all the 'important bits' that are exposed in HOCON configs.

 - It may be nice to provide a sort of 'forwarding' for providername with our own type/enum so users don't have to poke at `Linq2Db.ProviderName`
 - Validation/warning on invalid configs would also be a good idea (fail fast)

dkottis:
Fixes #18 

## Changes

Adds a DataOptionsSetup class to be able to configure custom DataOptions for [linq2db](https://github.com/linq2db/linq2db)

## Remarks

* I didn’t want to add whole new PersistenceSetup class, so it’s just a Setup for the DataOptions. Otherwise, there would be three ways to configure the plugin, and more testing would be necessary.

* If the data options are provided, the user must add the provider and the connection string again (if needed). Linq2Db checks for different configuration variations, and always adding the connection string and the provider name would exclude some configurations (like a connection factory).
[See here ](https://github.com/linq2db/linq2db/blob/89b1b78041d88cfe03829def11d9ff71870082de/Source/LinqToDB/Data/DataConnection.Configuration.cs#L501)
  
* In this first version I didn’t include a Multi Setup class, like AzureTableStorageMultiJournalSetup. Would this be a blocker to a PR?

* I didn’t include a test for this setup, because the configs are patched in the corresponding Journal/Snapshot classes. So I really don’t know where or how to put it. Do you have an advise on this?

* The `Copy` Methods and private constructors [like here](https://github.com/Astasian/Akka.Persistence.Sql/blob/1c4eeeb3792a52180552a45458a014bd3389167c/src/Akka.Persistence.Sql/Config/SnapshotConfig.cs#L108) could be replaced by records with `with` copying. But I didn't want to make any potential breaking changes.

## Checklist

* [x] Changes in public API reviewed, if any.
* [x] I have added website documentation for this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,10 +73,9 @@ var host = new HostBuilder()
         });
     })
 ```
-
-You must always provide a `connectionString` and a `providerName`, even if you provide a `DataOptions` object. 
-Please consult the Linq2Db documentation for more information on how to configure a valid `DataOptions` object. 
-The `MappingSchema` will always be overridden by Akka.Persistence.Sql.
+If `dataOptions` are provided, you must give enough information for linq2db to connect to the database.
+This includes setting the connection string and provider name again, if needed in your use case. 
+Please consult the Linq2Db documentation for more information on how to configure a valid `DataOptions` object.  `MappingSchema` and `RetryPolicy` will always be overridden by Akka.Persistence.Sql.
 
 ## The Classic Way, Using HOCON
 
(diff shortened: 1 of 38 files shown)


Aaronontheweb (approved):
LGTM - went through this over a call with @Arkatufus

PR #489, opened 2024-10-22
Fix SqlServer table initialization race condition

Arkatufus:
Using SqlServer, when several read/write journals of the same id were created at the same time, there is a possible race condition where 2 journals would try and create the same index for the same table at the same time, resulting in an exception:

```
Failure during [akka://test-sys/system/akka.persistence.journal.sharding#237109392] initialization.
Microsoft.Data.SqlClient.SqlException (0x80131904): The operation failed because an index or statistics with name 'IX_tags_persistence_id_sequence_nr' already exists on table 'tags'.
```

## Changes

* Add a TRY...CATCH guard in the T-SQL statements and discard any "index already exist" and "constraint already exist" errors.


Aaronontheweb (approved):
LGTM

PR #492, opened 2024-11-15
Include deleteCompatibilityMode setting explanation to migration guide

Arkatufus:


docs/articles/migration.md
```diff
@@ -219,3 +220,13 @@ To migrate your database to use the new WriterUuid feature, follow these steps:
    akka.persistence.journal.sql.{mapping-name}.journal.use-writer-uuid-column = true
    ```
 5. Bring the cluster back up.
+
+# Disable Delete Compatibility Mode (Optional)
+
+When you set `deleteCompatibilityMode` to `true`, `Akka.Persistence.Sql` will use the old `Akka.Persistence.Sql.Common` journal entry deletion implementation that is slower compared to the new `Akka.Persistence.Sql` implementation. Setting this to `false` will improve performance, however, manually migrating the old implementation table entries to the new one is error-prone and can cause data corruption, so we highly discourage you to do this.
+
+There is an edge case where you can set `deleteCompatibilityMode` to `false`, that is if you've **never** deletes anything from your journal table. Technically, the way you can make sure that you can set `deleteCompatibilityMode` to `false` and be able to delete the journal metadata table is by making sure that your use case satisfies the following criteria:
```
robindv: deletes -> deleted

Aaronontheweb: Done

PR #503, opened 2025-01-06
compare unadapted event count to batch size when determining if there are any more events to read

joerosenberg:
Fixes #
https://github.com/akkadotnet/Akka.Persistence.Sql/issues/502

## Changes

Please provide a brief description of the changes here.

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks

### DIFF
--- a/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
+++ b/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
@@ -226,7 +226,7 @@ namespace Akka.Persistence.Sql.Query
 
             return _readJournalDao
                 .Events(offset, latestOrdering.Max, max)
-                .AlsoToMaterialized(Sink.Aggregate(0L, (count, _) => count + 1), Keep.Right)
+                .AlsoToMaterialized(Sink.Aggregate<Try<(IPersistentRepresentation, string[], long)>, long>(0L, (readCount, _) => readCount + 1), Keep.Right)
                 .SelectAsync(1, r => Task.FromResult(r.Get()))
                 .SelectMany(
                     a =>
@@ -367,15 +367,14 @@ namespace Akka.Persistence.Sql.Query
                                     GetMaxOrderingId.Instance,
                                     askTimeout);
 
-                            var (xsTask, readCountTask) = CurrentJournalEvents(uf.offset, batchSize, queryUntil)
+                            var readBatch = CurrentJournalEvents(uf.offset, batchSize, queryUntil)
                                 .ToMaterialized(Sink.Seq<EventEnvelope>(), Keep.Both)
                                 .Run(_mat);
 
-                            await Task.WhenAll(xsTask, readCountTask);
-                            var xs = xsTask.Result;
-                            var readCount = readCountTask.Result;
+                            var rawReadCount = await readBatch.Item1;
+                            var xs = await readBatch.Item2;
 
-                            var hasMoreEvents = readCount == batchSize;
+                            var hasMoreEvents = rawReadCount == batchSize;
 
                             var nextControl = FlowControlEnum.Unknown;
                             if (terminateAfterOffset.HasValue)
--- a/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
+++ b/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
@@ -219,13 +219,14 @@ namespace Akka.Persistence.Sql.Query
                             timestamp: r.representation.Timestamp, 
                             tags: r.tags));
 
-        private Source<EventEnvelope, NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
+        private Source<EventEnvelope, Task<long>> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
         {
             if (latestOrdering.Max < offset)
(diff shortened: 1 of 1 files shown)


src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -219,13 +219,14 @@ private Source<EventEnvelope, NotUsed> EventsByPersistenceIdSource(
...
-        private Source<EventEnvelope, NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
+        private Source<EventEnvelope, Task<long>> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
         {
             if (latestOrdering.Max < offset)
-                return Source.Empty<EventEnvelope>();
+                return Source.Empty<EventEnvelope>().MapMaterializedValue(_ => Task.FromResult(0L));
 
             return _readJournalDao
                 .Events(offset, latestOrdering.Max, max)
+                .AlsoToMaterialized(Sink.Aggregate<Try<(IPersistentRepresentation, string[], long)>, long>(0L, (readCount, _) => readCount + 1), Keep.Right)
```
to11mtm, line 229: Build error here may be a langver issue....

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -366,10 +367,14 @@ private Source<EventEnvelope, NotUsed> Events(long offset, long? terminateAfterO
...
-                                .RunWith(Sink.Seq<EventEnvelope>(), _mat);
+                            var readBatch = CurrentJournalEvents(uf.offset, batchSize, queryUntil)
+                                .ToMaterialized(Sink.Seq<EventEnvelope>(), Keep.Both)
+                                .Run(_mat);
 
-                            var hasMoreEvents = xs.Count == batchSize;
+                            var rawReadCount = await readBatch.Item1;
+                            var xs = await readBatch.Item2;
+
+                            var hasMoreEvents = rawReadCount == batchSize;
```
to11mtm, line 377: Looking at this from an 'abstract' point of view, I'll ask the question: Could we just have `CurrentJournalEvents` refactort to have a `private ??? CurrentJournalEventsImpl` that returns both the batch and the raw rowcount as well as the events itself? 

May be uglier and cost more memory but *might* make logic easier to refactor, at least for first phase...

-----

Should also note, TBF I don't have confidence in the use of streams in this loop; it it 'as-ported' but I question whether it's the best call; at one point I looked into swapping much of this logic to *not* use streams for the inner loops. At the time, it was a wash, but `AlsoToMaterialized` with a `Keep.Both` may change the scales on that...

CC @Arkatufus  b/c this is another case where benchmarks on Read journal vs main journal would be useful.

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -219,13 +219,14 @@ private Source<EventEnvelope, NotUsed> EventsByPersistenceIdSource(
                             timestamp: r.representation.Timestamp, 
                             tags: r.tags));
 
-        private Source<EventEnvelope, NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
+        private Source<EventEnvelope, Task<long>> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
         {
             if (latestOrdering.Max < offset)
-                return Source.Empty<EventEnvelope>();
+                return Source.Empty<EventEnvelope>().MapMaterializedValue(_ => Task.FromResult(0L));
```
to11mtm, line 225: If this is how we finally roll (see other comments) I wonder whether or not having our `Task.FromResult(0L)` as a static for older runtimes is a good idea or not, at least for consistency on our side.

(OTOH, I forget if FW runtimes cache `Task.FromResult<long>(0)`... might be overthinking this...)

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
to11mtm, line 229: .... I can't help but ask if we should just enrich `Events` return flow here instead, relying on this aggregate doesn't look easy to comprehend/maintain in my head.

CC @Arkatufus + @Aaronontheweb , if we don't ever need equality on the value of these `Try` bits, it can be better from AOT perspective to just declare our own struct since then it doesn't have to generate generic equality code for the type.

to11mtm (commented):
Left some comments.

Should also note, if these changes are needed for the non-tagged Event queries, probably need to also also have changes on the 'by-tag' methods or at least some verification that those do not have problems in/around accidentally hammering the DB or doing something else weird/bad if the count is off.

Arkatufus:
Overall, I think this solution is overly complicated to address the issue. We can make the fix much simpler.

Arkatufus:
See #505 for a simpler solution

Aaronontheweb:
@joerosenberg sorry for the delay in merging this in, but we just merged https://github.com/akkadotnet/Akka.Persistence.Sql/pull/505 which should resolve the issue.

We would not have found this bug nor the fix for it were it not for your tremendous investigative work here - we deeply appreciate your effort, your code, and your detailed explanation that lead to this getting fixed. So even though we didn't merge your PR, your effort was still essential to discovering this issue and fixing it. We appreciate it.

PR #505, opened 2025-01-08
Fix query batch size detection failure

### LINKED ISSUE #502: `SqlReadJournal.CurrentAllEvents` completes before emitting all events when used with an `IReadEventAdapter` that drops events
**Version Information**
Version of Akka.NET? 1.5.30

**Describe the bug**
`SqlReadJournal.CurrentAllEvents` completes before emitting all events when used with an `IReadEventAdapter` that drops events (i.e. by returning `EventSequence.Empty`).

This is due to `hasMoreEvents` in `SqlReadJournal.Events` being set to false when `CurrentJournalEvents` returns a number of events not equal to the max batch size. Since `CurrentJournalEvents` returns adapted events, this can be a false positive if the event adapter does not perform a 1-to-1 mapping.

**To Reproduce**
1. Configure an `IReadEventAdapter` that returns `EventSequence.Empty` for some events
2. Perform a `CurrentAllEvents` query from offset 0 against a journal that a) has more events than the configured batch size (believe this is 500 by default), and b) contains an event in the first batch that will be mapped to `EventSequence.Empty` by the read adapter
3. Observe that the `CurrentAllEvents` query is completed after reading the batch containing the dropped event

**Actual behaviour**
See above

**Expected behaviour**
`CurrentAllEvents` should only complete after reaching end of journal

**Environment**
Windows, .NET 8

**Additional context** https://github.com/akkadotnet/Akka.Persistence.Sql/blob/fff29d16a2ccca2c4be2e88643d77fc7391b644c/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs#L366

https://github.com/akkadotnet/Akka.Persistence.Sql/blob/fff29d16a2ccca2c4be2e88643d77fc7391b644c/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs#L228

Arkatufus:
Fixes #502

An alternate way to #503 for fixing #502

## Changes

* Make `CurrentJournalEvents` to return a stream or `EventEnvelope` array, this way we can still count the actual number of rows returned by the database.
* Add regression spec

## Checklist

### DIFF
--- a/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
+++ b/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
@@ -219,15 +219,15 @@ namespace Akka.Persistence.Sql.Query
                             timestamp: r.representation.Timestamp, 
                             tags: r.tags));
 
-        private Source<EventEnvelope, NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
+        private Source<EventEnvelope[], NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
         {
             if (latestOrdering.Max < offset)
-                return Source.Empty<EventEnvelope>();
+                return Source.Empty<EventEnvelope[]>();
 
             return _readJournalDao
                 .Events(offset, latestOrdering.Max, max)
                 .SelectAsync(1, r => Task.FromResult(r.Get()))
-                .SelectMany(
+                .Select(
                     a =>
                     {
                         var (representation, tags, ordering) = a;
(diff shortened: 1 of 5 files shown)


src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -219,15 +219,15 @@ private Source<EventEnvelope, NotUsed> EventsByPersistenceIdSource(
                             timestamp: r.representation.Timestamp, 
                             tags: r.tags));
 
-        private Source<EventEnvelope, NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
+        private Source<EventEnvelope[], NotUsed> CurrentJournalEvents(long offset, long max, MaxOrderingId latestOrdering)
```
Arkatufus, line 222: Instead of using `SelectMany` and return a list of `EventEnvelope`s, we use `Select` and return a list of `EventEnvelope[]` arrays instead. This list count will be consistently equals to the number of rows we fetched from the database.

Arkatufus (commented):
Self-review

Aaronontheweb (approved):
LGTM

to11mtm (commented):
Beautiful. <3

PR #512, opened 2025-01-23
implement dual targetting for both netstandard2.0 and net60

Arkatufus:


Aaronontheweb:
Good idea, brings this up to par with the core Akka.NET library itself.

Aaronontheweb (approved):
LGTM

PR #518, opened 2025-02-10
Fix memory leak caused by QueryThrottler

### LINKED ISSUE #516: Memory Leak after updating Akka from 1.5.27.1 to 1.5.37
**Version Information**
Version of Akka.NET?
**1.5.37**
Which Akka.NET Modules?
**Akka.NET + Persistence, Remoting, Clustering. Journal Plugin is Akka.Persistence.Sql**

**Describe the bug**
Last week I updated our staging environment from Akka 1.5.27.1 to 1.5.37.
Spefically, I updated the following packages:
- Akka: 1.5.27.1 -> 1.5.37
- Akka.Cluster.Sharding: 1.5.27.1 -> 1.5.37
- Akka.Persistence.Sql: 1.5.25 -> 1.5.37
- Akka.Serialization.Hyperion: 1.5.27.1 -> 1.5.37

As part of this update, I also retargeted from .NET 7 to .NET 8.

After a few days, I noticed our staging environment running low on memory:

Image

Inspecting the VMs reveal most of processes hosting Akka.NET were over 1GB in size.
The only exceptions were some processes that do not instantiate any journal reads, which were not affected. This may mean that the memory leak is coming from Akka.Persistence or the Akka.Persistence.Sql plugin.

Additionally, I took a memory dump of one of the affected processes:

Image

Image

Image

`Task<QueryStartGranted>` seems to be the culprit, which further indicates that the memory leak is coming from Akka.Persistence module or its plugin.

**Environment**
Windows on .NET 8

**Additional context**
I am currently reverting our environment to 1.5.27.1 to verify that the memory leak disappears (just to confirm that this is not due to some other changes). Potentially, it could also be due to the upgrade to .NET 8, but probably less likely. I will monitor RAM usage and check back tomorrow to see if just reverting the packages version rollbacks the issue.

Let me know if there is anything else I can help with to narrow this down.

Arkatufus:
Fixes #516

## Changes

Make sure we don't death watch Ask temp actors 

## Checklist

### DIFF
--- a/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
+++ b/src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
@@ -52,14 +52,9 @@ namespace Akka.Persistence.Sql.Extensions
             DbStateHolder state,
             Func<AkkaDataConnection, CancellationToken, Task<T>> handler)
         {
-            using var cts = CancellationTokenSource.CreateLinkedTokenSource(state.ShutdownToken);
-            {
-                cts.CancelAfter(state.QueryThrottleTimeout);
-                await state.QueryPermitter.Ask<QueryStartGranted>(RequestQueryStart.Instance, cts.Token);
-            }
-            
             try
             {
+                await state.QueryPermitter.Ask<QueryStartGranted>(new RequestQueryStart(state.QueryThrottleTimeout), state.QueryThrottleTimeout);
                 return await factory.ExecuteWithTransactionAsync(state.IsolationLevel, state.ShutdownToken, handler);
             }
             finally
@@ -103,14 +98,9 @@ namespace Akka.Persistence.Sql.Extensions
             TState state,
             Func<AkkaDataConnection, CancellationToken, TState, Task<T>> handler)
         {
-            using var cts = CancellationTokenSource.CreateLinkedTokenSource(factory.ShutdownToken);
-            {
-                cts.CancelAfter(factory.QueryThrottleTimeout);
-                await factory.QueryPermitter.Ask<QueryStartGranted>(RequestQueryStart.Instance, cts.Token);
-            }
-            
             try
             {
+                await factory.QueryPermitter.Ask<QueryStartGranted>(new RequestQueryStart(factory.QueryThrottleTimeout), factory.QueryThrottleTimeout);
                 return await factory.ConnectionFactory.ExecuteWithTransactionAsync(state, factory.IsolationLevel, factory.ShutdownToken, handler);
             }
             finally
--- a/src/Akka.Persistence.Sql/Query/QueryThrottler.cs
+++ b/src/Akka.Persistence.Sql/Query/QueryThrottler.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Akka.Actor;
 using Akka.Event;
 using Akka.Pattern;
@@ -18,8 +19,26 @@ namespace Akka.Persistence.Sql.Query;
 /// </summary>
 internal sealed class RequestQueryStart
 {
-    public static readonly RequestQueryStart Instance = new();
-    private RequestQueryStart() { }
+    public DateTime DeadlineTime { get; }
+
+    public RequestQueryStart(TimeSpan timeout)
+    {
+        DeadlineTime = DateTime.UtcNow.Add(timeout);
+    }
+}
+
+internal sealed class PendingRequest
+{
+    public PendingRequest(IActorRef requester, DateTime deadlineTime)
+    {
+        Requester = requester;
+        DeadlineTime = deadlineTime;
+    }
+    
+    public IActorRef Requester { get; }
+    public DateTime DeadlineTime { get; }
+    
+    public bool IsExpired => DeadlineTime < DateTime.UtcNow;
 }
 
 /// <summary>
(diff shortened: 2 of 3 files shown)


src/Akka.Persistence.Sql/Extensions/ConnectionFactoryExtensions.cs
```diff
@@ -52,14 +52,9 @@ internal static async Task<T> ExecuteQueryWithTransactionAsync<T>(
             DbStateHolder state,
             Func<AkkaDataConnection, CancellationToken, Task<T>> handler)
         {
-            using var cts = CancellationTokenSource.CreateLinkedTokenSource(state.ShutdownToken);
-            {
-                cts.CancelAfter(state.QueryThrottleTimeout);
-                await state.QueryPermitter.Ask<QueryStartGranted>(RequestQueryStart.Instance, cts.Token);
-            }
```
Arkatufus, line 59: Remove complicated cancellation token code, replaced it with `Ask` internal timeout mechanism

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
```diff
@@ -40,31 +59,72 @@ internal sealed class ReturnQueryStart
...
         
-        Receive<RequestQueryStart>(_ =>
+        Receive<RequestQueryStart>(request =>
         {
-            Context.Watch(Sender);
+            if(Sender is ActorRefWithCell)
+            {
+                _watchCount++;
+                Context.Watch(Sender);
+            }
```
Arkatufus, line 124: Actual fix, never watch any `IActorRef` with no `ActorCell`, these actors are not compatible with death watch.

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
```diff
@@ -95,19 +173,29 @@ private void QueryStartGranted(IActorRef actorRef)
...
+        }
 
-        var popRef = _pending.First?.Value;
-        if (popRef is not null)
+        while (_pending.First is not null)
         {
+            var pending = _pending.First.Value;
             _pending.RemoveFirst();
-            QueryStartGranted(popRef);
+            if (pending is not null && pending.IsExpired)
```
Arkatufus, line 201: Since we can't rely on death watch mechanism to automatically purge pending requests, we have to add timeout deadlines to the pending requests. All expired pending requests will be discarded.

Arkatufus (commented):
Self-review

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
```diff
@@ -40,31 +59,72 @@ internal sealed class ReturnQueryStart
...
+#endregion
+
 /// <summary>
 /// Token bucket throttler that grants queries permissions to run each iteration
 /// </summary>
 /// <remarks>
-/// Works identically to the RecoveryPermitter built into Akka.Persistence.
+/// Works almost identically to the RecoveryPermitter built into Akka.Persistence.
+/// 
+/// NOTE: Since this permitter works with Ask and we can't death watch an Ask temp
```
to11mtm: It would be nice if we could encapsulate/sugar this behavior in future, that said comment is great breadcrumb for future improvement.

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
```diff
@@ -95,19 +173,29 @@ private void QueryStartGranted(IActorRef actorRef)
...
-    private void ReturnQueryPermit(IActorRef actorRef)
+    private void ReturnQueryPermit()
     {
         _usedPermits--;
-        Context.Unwatch(actorRef);
 
         if (_usedPermits < 0)
-            throw new IllegalStateException("Permits must not be negative");
+        {
+            _log.Warning("Permits must not be negative");
```
to11mtm, line 192: Is it overkill to include the negative value here? Also what justifies the reset to zero?

to11mtm (commented):
left a few questions/comments, great work digging into it. At most it's future improvement in my head...

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
Arkatufus: Its a warning for the developers. It should not kill the actor outright, and the throttler should keep on going. If the value goes negative, this means that we over-return permits, something is returning more permits than what was issued. Its not a critical error, the actor should work regardless, but we should take a note of it happening.

to11mtm: Makes sense; Should we add a comment in code AFA your comment? Mostly asking in case other maintainers/devs ask same question in future...

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
```diff
@@ -40,31 +59,72 @@ internal sealed class ReturnQueryStart
...
+///       actor, there is a possibility that a permit requester failed to return
+///       the request permit.
+/// 
+///       ALWAYS USE A TRY...FINALLY BLOCK WHEN USING ASK AND RETURN THE PERMIT IN
+///       THE FINALLY BLOCK
 /// </remarks>
 internal sealed class QueryThrottler : ReceiveActor
 {
-    private readonly LinkedList<IActorRef> _pending = new();
+    private readonly LinkedList<PendingRequest> _pending = new();
```
to11mtm, line 108: .... The use of LinkedList here is curious, if nothing else I wonder if an UnrolledLinkedList or similar would be worth considering (I'm guessing we use this for performance?)

Arkatufus: Its mostly because the .NET linked list is a doubly linked list implementation, the read/write time is constant for the first and last element (O(1)). It is just the perfect fit for the implementation.

Aaronontheweb (approved):
LGTM - we need to get rid of `Context.Watch` here even with https://github.com/akkadotnet/akka.net/pull/7502 because now the query permit gets returned before the query even runs, which makes the original throttling implementation ineffective.

src/Akka.Persistence.Sql/Query/QueryThrottler.cs
Arkatufus: good idea

Issue #530, opened 2025-04-06
Multiple ReadJournal instances cause actor name conflicts and incorrect query throttling


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.5.40
Which Akka.NET Modules? Akka.Persistence.Sql 

## Problem Description
When running multiple persistence journals in the same Akka.NET application with Akka.Persistence.Sql, creating multiple ReadJournal instances causes actor name conflicts. This occurs when multiple projection actors (ReceivePersistentActor instances) try to create their own read journals using `Context.System.ReadJournalFor<SqlReadJournal>()`.

### Error Messages
Users encounter either of these exceptions:
- `Actor name "akka.persistence.query.journal.contact-query-permitter" is not unique\!`
- `Actor name "journalakka-persistence-sql-sequence-actor" is not unique\!`

### Current Implementation Issues
1. ReadJournal is being created multiple times unnecessarily
2. Each ReadJournal creates its own query throttler, defeating the purpose of having a throttler in the first place
3. The actors are created in the /user hierarchy instead of being /system actors

### Reproduction Steps
1. Create multiple persistence journals in the same Akka.NET application
2. Use these journals for data projections to various destinations (SQL, RabbitMQ, etc.)
3. Run multiple projector actors simultaneously
4. When the second projection actor reaches its OnReplaySuccess() method and calls:
```csharp
var readJournal = Context.System.ReadJournalFor<SqlReadJournal>("my-domain-name");
```
The exception is thrown.

### Technical Details
- **Version**: 1.5.38.1 (also present in 1.5.40)
- **Relevant Code**:
  - [SqlReadJournal.cs\#L86-L88](https://github.com/akkadotnet/Akka.Persistence.Sql/blob/f8b7a68f5de4d1996738e8c5ed011c22089a083c/src/Akka.Persistence.Sql/Query/SqlReadJournal.cs#L86-L88)
  - [SqlReadJournalProvider.cs\#L25-L26](https://github.com/akkadotnet/Akka.Persistence.Sql/blob/6cc080dc34e3011c6715c36b4198fad7127282ec/src/Akka.Persistence.Sql/Query/SqlReadJournalProvider.cs#L25-L26)

## Proposed Solutions
1. Make the ReadJournal and query throttler singleton instances
2. Move the actors to the /system hierarchy instead of /user
3. Implement proper naming/scoping for these actors when multiple journals are in use

## Impact
[...]

PR #531, opened 2025-04-07
Bugfix #530

Arkatufus:
## Changes

* Add unit tests
* Refactor all internal actors to system actors
* Add bugfix

### DIFF
--- a/src/Akka.Persistence.Sql.Tests/Query/Issue530Spec.cs
+++ b/src/Akka.Persistence.Sql.Tests/Query/Issue530Spec.cs
@@ -1,0 +1,82 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="Issue530Spec.cs" company="Akka.NET Project">
+//      Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Hosting;
+using Akka.Persistence.Query;
+using Akka.Persistence.Sql.Query;
+using Akka.Persistence.Sql.Tests.Common.Containers;
+using Akka.Persistence.Sql.Tests.Sqlite;
+using Akka.Persistence.TCK;
+using FluentAssertions;
+using LanguageExt.UnitsOfMeasure;
+using Xunit;
+using Xunit.Abstractions;
+using static FluentAssertions.FluentActions;
+
+namespace Akka.Persistence.Sql.Tests;
+
+[Collection(nameof(SqlitePersistenceSpec))]
+public class Issue530Spec: PluginSpec
+{
+    public Issue530Spec(ITestOutputHelper output, SqliteContainer fixture)
+        : base(FromConfig(Config(fixture)), nameof(Issue530Spec), output)
+    {
+    }
+    
+    [Fact]
+    public void MultipleReadJournalsTest()
+    {
+        var query = PersistenceQuery.Get(Sys);
+        
+        Invoking(() =>
+            {
+                var readJournal1 = query.ReadJournalFor<SqlReadJournal>("akka.persistence.query.journal.sql");
+                var readJournal2 = query.ReadJournalFor<SqlReadJournal>("akka.persistence.query.journal.sql2");
+            })
+            .Should().NotThrow<InvalidActorNameException>();
+    }
+
+    private static Configuration.Config Config(SqliteContainer fixture)
+        {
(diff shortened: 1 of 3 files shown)


src/Akka.Persistence.Sql.Tests/Query/Issue530Spec.cs
```diff
@@ -0,0 +1,82 @@
...
+
+[Collection(nameof(SqlitePersistenceSpec))]
+public class Issue530Spec: PluginSpec
+{
+    public Issue530Spec(ITestOutputHelper output, SqliteContainer fixture)
+        : base(FromConfig(Config(fixture)), nameof(Issue530Spec), output)
+    {
+    }
+    
+    [Fact]
```
Aaronontheweb, line 33: Please file a GitHub issue to get this added to the TCK for Akka.Persistence.Query.

Arkatufus: Done

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -101,12 +101,12 @@ public SqlReadJournal(
...
                 default);
 
-            _journalSequenceActor = system.ActorOf(
+            _journalSequenceActor = system.SystemActorOf(
                 props: Props.Create(
                     () => new JournalSequenceActor(
                         _readJournalDao,
                         _readJournalConfig.JournalSequenceRetrievalConfiguration)),
-                name: $"{_readJournalConfig.TableConfig.EventJournalTable.Name}akka-persistence-sql-sequence-actor");
+                name: $"{_readJournalConfig.PluginId}-{_readJournalConfig.TableConfig.EventJournalTable.Name}-akka-persistence-sql-sequence-actor");
```
Arkatufus, line 109: Actual fix. 

Original intent:
We assume that users will only use this plugin to connect to one and only one database instance, the event journal table name was thought as unique enough an identifier for a read journal sequence actor name.

Problem:
Since Akka.Persistence.Sql can target multiple databases, it is possible that each database will use the same event journal table name and the naming convention will cause an actor name collision.

> [!NOTE]
> Since we're using the read journal plugin id to make the actor name unique instead of the write journal plugin id, this does mean that we do not guarantee that there will only be a single read journal for a single write journal. User will need to be responsible for managing the number of read journal configurations and the read journal they instantiate to ensure system responsiveness.

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -83,7 +83,7 @@ public SqlReadJournal(
                 namePrefix: $"l2db-query-mat-{Guid.NewGuid():N}");
 
             _log = Logging.GetLogger(system, $"{_readJournalConfig.PluginId}-{nameof(SqlReadJournal)}");
-            _queryPermitter = system.ActorOf(
+            _queryPermitter = system.SystemActorOf(
```
Arkatufus, line 86: Refactor all internal actors to system actors

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
```diff
@@ -101,12 +101,12 @@ public SqlReadJournal(
                 // TODO: figure out a way to signal shutdown to the query executor here
                 default);
 
-            _journalSequenceActor = system.ActorOf(
+            _journalSequenceActor = system.SystemActorOf(
```
Arkatufus, line 104: Refactor all internal actors to system actors

src/Akka.Persistence.Sql/Query/SqlReadJournalProvider.cs
```diff
@@ -22,6 +23,16 @@ public SqlReadJournalProvider(
...
+        /// <summary>
+        ///     Note that this is safe to do because the only place this is being called is
+        ///     inside the `PersistenceQuery.ReadJournalFor{T}()` public method inside
+        ///     Akka.Persistence.Query and the result of that method call is then cached
+        ///     and reused for the duration of the ActorSystem lifetime.
+        /// </summary>
+        /// <returns>
+        ///     A new instance of IReadJournal specific to this persistence plugin.
+        /// </returns>
+        [InternalApi]
```
Arkatufus, line 35: Add explanation why this is safe to do

Arkatufus (commented):
self-review

src/Akka.Persistence.Sql/Query/SqlReadJournal.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM

PR #538, opened 2025-05-21
Improve journal delete performance

Arkatufus:
> [!NOTE]
>
> These changes are safe because we're using transactions, the old JVM implementation queries were designed that way because they have to account for possible non-transaction delete failures.
> We need transaction guarantee because we have to write to multiple tables to support backward compatibility and tag table indexing.

## Changes

Optimize event journal delete code
* Shorten the codes living inside the transaction block
* Optimize SQL query

### DIFF
--- a/src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
+++ b/src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
@@ -42,7 +42,7 @@ namespace Akka.Persistence.Sql.Db
 
             _useCloneDataConnection = config.UseCloneConnection;
 
-            if (_opts.RetryPolicyOptions.RetryPolicy is null && _opts.ConnectionOptions.ProviderName!.ToLowerInvariant().StartsWith("sqlserver"))
+            if (_opts.RetryPolicyOptions.RetryPolicy is null && _opts.RetryPolicyOptions.Factory is null && _opts.ConnectionOptions.ProviderName!.ToLowerInvariant().StartsWith("sqlserver"))
                 _opts = _opts.WithOptions( _opts.RetryPolicyOptions with { RetryPolicy = new SqlServerRetryPolicy() } );
             
             _cloneConnection = new Lazy<AkkaDataConnection>(
--- a/src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
@@ -119,6 +119,9 @@ namespace Akka.Persistence.Sql.Journal.Dao
             {
                 maxMarkedDeletion = await MaxMarkedForDeletionMaxPersistenceIdQuery(connection, persistenceId, maxSequenceNr).FirstOrDefaultAsync(ShutdownToken);
             }
+
+            if (maxMarkedDeletion is 0)
+                return;
             
             await ConnectionFactory.ExecuteWithTransactionAsync(
                 WriteIsolationLevel,
--- a/src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
+++ b/src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
@@ -112,22 +112,34 @@ namespace Akka.Persistence.Sql.Journal.Dao
 
         public async Task Delete(string persistenceId, long maxSequenceNr)
         {
+            long maxMarkedDeletion;
+
+            // no need to use transaction
+            await using (var connection = ConnectionFactory.GetConnection())
+            {
+                maxMarkedDeletion = await MaxMarkedForDeletionMaxPersistenceIdQuery(connection, persistenceId, maxSequenceNr).FirstOrDefaultAsync(ShutdownToken);
+            }
+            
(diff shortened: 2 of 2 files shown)


src/Akka.Persistence.Sql/Db/AkkaPersistenceDataConnectionFactory.cs
```diff
@@ -42,7 +42,7 @@ public AkkaPersistenceDataConnectionFactory(IProviderConfig<JournalTableConfig>
 
             _useCloneDataConnection = config.UseCloneConnection;
 
-            if (_opts.RetryPolicyOptions.RetryPolicy is null && _opts.ConnectionOptions.ProviderName!.ToLowerInvariant().StartsWith("sqlserver"))
+            if (_opts.RetryPolicyOptions.RetryPolicy is null && _opts.RetryPolicyOptions.Factory is null && _opts.ConnectionOptions.ProviderName!.ToLowerInvariant().StartsWith("sqlserver"))
```
Arkatufus, line 45: Add missing retry policy factory check. The `Factory` (if not null) is used internally by Linq2Db to populate `DataConnection.RetryPolicy` if the `DataOption.RetryPolicyOptions.RetryPolicy` is null.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -112,22 +112,37 @@ public async Task<IImmutableList<Exception>> AsyncWriteMessages(
 
         public async Task Delete(string persistenceId, long maxSequenceNr)
         {
+            long maxMarkedDeletion;
+
+            // no need to use transaction
+            await using (var connection = ConnectionFactory.GetConnection())
+            {
+                maxMarkedDeletion = await MaxMarkedForDeletionMaxPersistenceIdQuery(connection, persistenceId, maxSequenceNr).FirstOrDefaultAsync(ShutdownToken);
+            }
```
Arkatufus: Retrieve the highest affected sequence number that is affected by the delete operation. No need to use transaction, this is strictly a read operation.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -112,22 +112,37 @@ public async Task<IImmutableList<Exception>> AsyncWriteMessages(
...
+
+            if (maxMarkedDeletion is 0)
+                return;
```
Arkatufus, line 127: Early bail out condition, just return if there's nothing to delete.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -112,22 +112,37 @@ public async Task<IImmutableList<Exception>> AsyncWriteMessages(
...
                 {
-                    await connection
-                        .GetTable<JournalRow>()
+                    var journalTable = connection.GetTable<JournalRow>();
+                    await journalTable
                         .Where(
                             r =>
                                 r.PersistenceId == persistenceId &&
-                                r.SequenceNumber <= maxSequenceNr)
+                                r.SequenceNumber == maxMarkedDeletion)
```
Arkatufus, line 139: Instead of marking all of the records less than the requested sequence number as deleted, we mark the highest affected instead. Update 1 record instead of many.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -112,22 +112,37 @@ public async Task<IImmutableList<Exception>> AsyncWriteMessages(
...
                         .UpdateAsync(token);
 
-                    var maxMarkedDeletion =
-                        await MaxMarkedForDeletionMaxPersistenceIdQuery(connection, persistenceId).FirstOrDefaultAsync(token);
+                    await journalTable
+                        .Where(
+                            r =>
+                                r.PersistenceId == persistenceId &&
+                                r.SequenceNumber < maxMarkedDeletion)
+                        .DeleteAsync(token);
```
Arkatufus, line 148: Simplify the deletion query, just delete anything that is less than the highest affected row.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -436,10 +439,11 @@ protected static ImmutableList<Exception> BuildWriteRejections(List<Util.Try<Jou
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static IQueryable<long> MaxMarkedForDeletionMaxPersistenceIdQuery(
             AkkaDataConnection connection,
-            string persistenceId)
+            string persistenceId, 
+            long maxSequenceNr)
             => connection
                 .GetTable<JournalRow>()
-                .Where(r => r.PersistenceId == persistenceId && r.Deleted)
+                .Where(r => r.PersistenceId == persistenceId && r.SequenceNumber <= maxSequenceNr)
```
Arkatufus, line 452: Optimization, instead of checking against persistence id and deleted field, check against persistence id and sequence number instead, as they are indexed.

Arkatufus (commented):
Self-review

to11mtm (approved):
Looks OK to me

PR #545, opened 2025-08-16
update to 6.0

to11mtm:
Fixes #

## Changes

Please provide a brief description of the changes here.

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks


src/Akka.Persistence.Sql.HelperLib/TagTableMigrator.cs
```diff
@@ -14,6 +14,7 @@
 using Akka.Persistence.Sql.Db;
 using Akka.Persistence.Sql.Journal.Types;
 using LinqToDB;
+using LinqToDB.Async;
```
to11mtm, line 17: Things got moved, now this namespace is a thing for `ToListAsync`/`ToDictionaryAsync` etc.

src/Akka.Persistence.Sql.TagTableMigration/Options.cs
```diff
@@ -65,8 +65,8 @@ public static string ToHocon(this ProviderType type)
                 ProviderType.SqlServer2019 => ProviderName.SqlServer2019,
                 ProviderType.SqlServer2022 => ProviderName.SqlServer2022,
                 ProviderType.MySql => ProviderName.MySql,
-                ProviderType.MySqlOfficial => ProviderName.MySqlOfficial,
-                ProviderType.MySqlConnector => ProviderName.MySqlConnector,
+                ProviderType.MySqlOfficial => ProviderName.MySql80MySqlData,//.MySqlOfficial,
+                ProviderType.MySqlConnector => ProviderName.MySql80MySqlConnector,
```
to11mtm, line 69: Should we add `ProviderType` changes here for the Tag Table migration Big difference is things are now a bit more explicit for MySQL versions...

Not sure if this implies config updates required (We could always try to translate if needed I guess?

src/Akka.Persistence.Sql.Tests.Common/Containers/MySqlContainer.cs
```diff
@@ -38,7 +38,7 @@ public MySqlContainer() : base("mysql", "8", $"mysql-{Guid.NewGuid():N}")
 
         public override string ConnectionString => _connectionStringBuilder.ToString();
 
-        public override string ProviderName => LinqToDB.ProviderName.MySqlOfficial;
+        public override string ProviderName => LinqToDB.ProviderName.MySql80MySqlConnector;//.MySqlOfficial;
```
to11mtm, line 41: Possibly should actually be `MySql80MySqlData` ?

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
```diff
@@ -43,7 +43,7 @@ public void Dispose()
             => _connection.Dispose();
 
         public AkkaDataConnection Clone()
-            => new(_providerName, (DataConnection)_connection.Clone());
+            => throw new NotSupportedException();//_providerName, (DataConnection)_connection.Clone());
```
to11mtm, line 46: We have ~~two~~three options here...

Option 1 - Take what I did here and treat it as Not Supported.

Option 2 - Have AkkaDataConnection take in the actual connection string as well, and then just have `Clone()` use that 

Option 3 - Just get rid of this method; -should- be a compile time break (bad) but should show up right away in most sane CI setups.

(If we go with 1 or 2, we'll still need to mark this as obsolete via attr)

to11mtm (commented):
Left some comments, a couple questions do come up...

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
Arkatufus: I think we should try to preserve backward compatibility for at least a few minor versions.
I personally think option 2 with obsolete attribute for this update, then changed to obsolete attribute that throws an error on later versions. We can then remove this API on later versions.

What do you think, @to11mtm and @Aaronontheweb?

src/Akka.Persistence.Sql.TagTableMigration/Options.cs
Arkatufus: Either we add all of the new `ProviderName` constant names and remove the missing ones, or we just have the user provide the proper string value for their target provider and remove `ProviderType` altogether. 

This isn't really sensitive to breaking change as it is only used as a rather simple string argument to enum to const transform, it is not used anywhere else.

src/Akka.Persistence.Sql.Tests.Common/Containers/MySqlContainer.cs
Arkatufus: The sql client package we're using for unit test is the MySql.Data, so I think MySql80MySqlData is the correct const here. We are using the MySql 8.0 container for testing, so the version is correct.

to11mtm:
Closed in favor of #580 which includes feedback from here. (or should anyway)

PR #580, opened 2026-03-20
Upgradde to Linq2Db 6 with Basic `.AsQueryable()` tag insert enhancement.

to11mtm:
### Feature 1:

Applies `.AsQueryable` branch to Insert pipeline such that:

 - Postgres
 - MSSQL
 - SQLite
 - Maybe MariaDB? (Didn't turn it on.)

Are able to use `.AsQueryable()` + `.InsertWithOutputAsync()`, to mitigate the performance issues around tagged event results in 'one event roundtrip, bulkinsert tags for that event', this will for supporting databases:

1. Build a lookup for post event insert for what to use on tag insert
2. Append entries to a collection and increment counters
3. When counters go above threshold, flush the entries and the lookups based on joining from `.InsertWithOutputAsync`.

Basically, it's a simplified version of how Linq2Db's `MultipleRows` `BulkCopy` works, we just have to do more to manually define the round trips.

---

### Feature 2:

Added `ChannelBatchQueue` Batcher and wired up into `BaseByteArrayJournalDao`

The main advantage of this setup, is that since the Queue read integrates the batching stage, we are able to avoid cases where `.SelectAsync` winds up pulling a single item only until all writer threads are busy...

Main added benefit is, this change actually improves the _existing_ tag write pipelines as well; since we are executing inside a transaction, even on the old path running more `single row->set of tags` inserts in a single transaction yields a performance benefit.

## Checklist

### Latest `dev` Benchmarks 

## Benchmarks

Added new benchmark tests for Postgres and MSSQL;

 - CSV Mode with 2 tags per event.
 - Tag table mode with current logic but actually has 2 tags per event
 - Tag table mode with AsQueryableInsert enabled, no tags on events
 - Tag table mode with AsQueryableInsert enabled, 2 tags on events.

**Removed prior benchmark results since now one would need to compare to dev.**


src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +342,177 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
                     }
                     else
                     {
-                        var config = JournalConfig.DaoConfig;
-                        var tail = xs;
-                        while (tail.Count > 0)
+                        if 
+                            (
+                                true && // Change to True to see performance impact of using fast insert on supported DBs.
```
to11mtm: ~~Per comments, I left this in to make it somewhat easier to compare with/without...~~

~~Probably could just be a config value too ? If nothing else lets it be opt-in and makes it somewhat easier to test....~~

Changed this to be a config knob, defaulted to FALSE.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +342,177 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+        }
+
+        protected async Task RunFastInsertNoEventParams(AkkaDataConnection connection, Seq<JournalRow> xs, BaseByteArrayJournalDaoConfig journalConfigDaoConfig, CancellationToken token)
+        {
+            // TODO: Should this be Configurable (including but not limited to 
+            // This version is also based on using roundTripByteLimit with a lazy heuristic to keep calculation logic simple.
+            
+            var roundTripByteLimit = 10_000_000;
+            var roundTripParamLimit = 999; // should be set based on provider tho.
+            var rowLimit = 200; // IDK
```
to11mtm: Per the comments this possibly could be fed from config...

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +342,177 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                    currBytes = 0;
+                }
+                else
+                {
+                    tagDict[(journalRow.PersistenceId, journalRow.SequenceNumber)] = journalRow.TagArray;
+                    currRows++;
+                    currBytes += 
+                        (journalRow.Message.Length * 2 // SqlText cost
+                            +
+                            256 // overparanoid cost for other fields on row.
```
to11mtm: This is just a stab in the dark, basically trying to estimate what is string cost of the non-message fields and other overhead.

Probably safe to be more conservative here TBH, e.x. 1024.

src/Akka.Persistence.Sql/Serialization/PersistentRepresentationSerializer.cs
```diff
@@ -103,7 +103,10 @@ public Try<T> Serialize(IPersistentRepresentation persistentRepresentation, long
             {
                 Tagged t => Serialize(persistentRepresentation.WithPayload(t.Payload), t.Tags, timeStamp),
 
-                _ => Serialize(persistentRepresentation, ImmutableHashSet<string>.Empty, timeStamp),
+                 // I'm a sloppy mechanic so this is the other part of demonstrating the pathological case, i.e. every event has tags.
+                
+                // _ => Serialize(persistentRepresentation, ImmutableHashSet<string>.Empty, timeStamp),
+                _ => Serialize(persistentRepresentation, ImmutableHashSet.Create("perf-test"), timeStamp),
```
to11mtm: Per comments, This is as it is at the moment to make it easier to test perf impact of tag cases with the benchmarks.

It should not exist in final PR, would love if we had a cleaner way to run benchmarks for tagtable cases with tags...

to11mtm (commented):
Added comments

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +350,167 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                            Identifier = jr.Identifier,
+                            WriterUuid = jr.WriterUuid
+                        }));
+                    await InsertJournalEntriesWithTags(connection, journalConfigDaoConfig, token, tagDict, query);
+                    insertList.Clear();
+                    tagDict.Clear();
+                    currRows = 0;
+                    currBytes = 0;
+                }
+                else
```
to11mtm: `else` here is bug I mentioned on discord

to11mtm: fixed

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +350,167 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+            var tagsToInsert = inserted.Join(
+                    tagDict,
+                    i => (i.PersistenceId, i.SequenceNumber),
+                    d => d.Key,
+                    (i, d) =>
+                        d.Value.Select(t => new JournalTagRow()
+                                { OrderingId = i.Ordering, PersistenceId = i.PersistenceId, SequenceNumber = i.SequenceNumber, TagValue = t })
+                            //.ToList()
+                )
+                .ToList();
```
to11mtm: So, FWIW we do this dance _because_ Certain databases do not guarantee output ordering on `INSERT WITH OUTPUT/RETURNING` (Actually I'm not sure if any do aside from mayyyyyybe Sqlite.) But otherwise an RDBMS is free to decide the order of an output statement, and from past research people say it is a real hazard on MSSQL at least.

And, it makes sense. Lots of RDBMS will try to start outputting as soon as possible, and depending on index pages etc the inserts may not always fire in the order expected.)

So, the simplest safe thing to do here, is to get back the PID and Seq (Cause those are supposed to be guaranteed by constraints for uniqueness) alongside the ordering.

##### All of that said, A fun question:

Should we consider doing `inserted.OrderBy(insRow => insRow.OrderingId)` before the join for additional data sanity? Sure this is all 'in a transaction' but I know there have been concerns about Query API and odd results in the past so wanted to at least note.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -55,21 +84,40 @@ protected BaseByteArrayJournalDao(
             Serializer = new ByteArrayJournalSerializer(config, serializer, config.PluginConfig.TagSeparator, selfUuid);
             _deserializeFlowMapped = Serializer.DeserializeFlow().Select(MessageWithBatchMapper());
             _tagWriteMode = JournalConfig.PluginConfig.TagMode;
+            _useTagTableAsQueryable = (JournalConfig.DaoConfig.UseTagTableAsQueryableLiteralInsert
+                                       &&
+                                       (JournalConfig.ProviderName.Contains("SqlServer") || JournalConfig.ProviderName.Contains("PostgreSQL") ||
+                                        JournalConfig.ProviderName.Contains("Sqlite")));
```
to11mtm: lots of config checks here so wanted to keep off hot path

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -55,21 +84,40 @@ protected BaseByteArrayJournalDao(
...
-                    r => new WriteQueueSet(ImmutableList.Create([r.Tcs]), r.Rows, ImmutableList.Create([r.CancellationToken])),
-                    (oldRows, newRows) =>
-                        new WriteQueueSet(
-                            oldRows.Tcs.Add(newRows.Tcs),
-                            oldRows.Rows.Concat(newRows.Rows),
-                            oldRows.CancellationTokens.Add(newRows.CancellationToken)))
+            _batcher = new ChannelQueueWithBatch<WriteQueueEntry, WriteQueueSet>(
+                _inputChannel.Reader,
+                maxWeight: 
+                // Sub 100 MaxWeight improves throughput for most cases...
```
to11mtm, line 127: I left this comment because it was a very clear observation from certain benchmarks, that at least with tag mode enabled, smaller batch sizes wind up faster.

Sometimes _far_ faster, pic related:

Image

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -268,34 +315,41 @@ public async Task<long> HighestSequenceNr(string persistenceId, long fromSequenc
...
-                    break;
-
-                case QueueOfferResult.QueueClosed:
+                            "Failed to enqueue journal row batch write, the queue was closed."));
+                else
                     promise.TrySetException(
                         new Exception(
-                            "Failed to enqueue journal row batch write, the queue was closed."));
-                    break;
+                            $"Failed to enqueue journal row batch write, the queue buffer was full ({JournalConfig.DaoConfig.BufferSize} elements)"));
```
to11mtm, line 375: So, nice thing about using a `BoundedChannel` here, is that we can _synchronously_ check whether we can write or not.

In general, seems that the tradeoff of the locks vs Akka streams machinery for Queue is a wash here...

Edited to add: by 'Here' I mean it doesn't make this part obviously faster....

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,209 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                    : JournalConfig.DaoConfig.BatchSize;
+            var currRows = 0;
+            var currBytes = 0;
+            var thisInsTagSize = 0;
+            Dictionary<(string PersistenceId, long SequenceNumber), string[]> tagDict 
+                = new Dictionary<(string persistenceId, long sequenceNumber), string[]>();
+            JournalRow[]? insertList = null;
+            try
+            {
+                insertList = JournalRowPool.Rent(rowLimit);
```
to11mtm, line 593: Felt like the responsible thing to do.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,199 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                            {
+                                PersistenceId = input.PersistenceId,
+                                SequenceNumber = input.SequenceNumber,
+                                Message = input.Message,
+                                Deleted = input.Deleted,
+                                Manifest = input.Manifest,
+                                Timestamp = input.Timestamp,
+                                Identifier = input.Identifier,
+                            },
+                        (inserted) => new { inserted.Ordering, inserted.PersistenceId, inserted.SequenceNumber }, token)
```
to11mtm, line 697: There's part of me that wanted to put these `Expression`s as statics and have switching logic etc to simplify IL here...

But, I know that is sometimes 'too much' 😅

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,199 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+            {
+                if (tagDict.TryGetValue((ir.PersistenceId, ir.SequenceNumber), out var tags))
+                {
+                    insertList.AddRange(tags.Select(t => new JournalTagRow()
+                    {
+                        OrderingId = ir.Ordering,
+                        PersistenceId = ir.PersistenceId,
+                        SequenceNumber = ir.SequenceNumber,
+                        TagValue = t
+                    }));
```
to11mtm: Could do a foreach here but I was 50/50 on it since we often see this pattern work just as well nowadays.

Speaking of which, yes the TryGetValue loop is because it did work better than `Join` linq op for local testing

src/Akka.Persistence.Sql/Journal/Dao/ChannelQueueWithBatch.md
```diff
@@ -0,0 +1,523 @@
+# Design for ChannelQueueWithBatch\<TInput, TBatch\>
```
to11mtm, line 1: I will of course remove this but at least wanted to have it out there, @Aaronontheweb would possibly be fun to run your setup against this design doc and PR to look at it more critically.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -55,21 +84,40 @@ protected BaseByteArrayJournalDao(
...
+                    r.Rows,
+                    ImmutableList.Create([r.CancellationToken])),
+                aggregate: (oldRows, newRows) =>
+                    new WriteQueueSet(
+                        oldRows.Tcs.Add(newRows.Tcs),
+                        oldRows.Rows.Concat(newRows.Rows),
+                        oldRows.CancellationTokens.Add(newRows.CancellationToken)));
+
+            Source.ChannelReader(_batcher)
+                .Buffer(JournalConfig.DaoConfig.Parallelism, OverflowStrategy.Backpressure)
```
to11mtm: The Buffer feels _fair_ for this overall conversion, and it seems to help numbers for legacy patterns and the like.

It might introduce more single writes, however since the parent batch is still aggressively pulling (Where old pattern could not) it seems to provide the most consistent performance against different batch size config values across scenarios.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -89,9 +137,8 @@ protected BaseByteArrayJournalDao(
                         return NotUsed.Instance;
                     })
                 .AddAttributes(ActorAttributes.CreateSupervisionStrategy(Deciders.RestartingDecider))
-                .ToMaterialized(
-                    Sink.Ignore<NotUsed>(),
-                    Keep.Left).Run(Materializer);
+                .To(Sink.Ignore<NotUsed>())
+                .Run(Materializer);
```
to11mtm, line 164: With this new setup, we have nothing to materialize.

Main half-concern I have, is whether this means we should be adding some forms of explicitly closing the input Channel. (I looked at what we did for `Source.Channel` and saw that we don't really close the input channel ever?)

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -268,34 +315,41 @@ public async Task<long> HighestSequenceNr(string persistenceId, long fromSequenc
...
+        /// <b>CopilotNote:</b> With <see cref="BoundedChannelFullMode.Wait"/>,
+        /// <c>TryWrite</c> is non-blocking and returns <c>false</c> when:
+        /// <list type="bullet">
+        ///   <item>Channel is at capacity (equivalent to old <c>Dropped</c>)</item>
+        ///   <item>Channel was completed with error (equivalent to old <c>Failure</c>)</item>
+        ///   <item>Channel was completed normally (equivalent to old <c>QueueClosed</c>)</item>
+        /// </list>
+        /// We distinguish these cases via <c>_batcher.Completion</c>.
+        /// </para>
+        /// </summary>
```
to11mtm, line 357: My custom copilot setup is _so excited_ about this that my rules for how to write xmldocs were ignored.

Just saying 👼

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+        {
+            public string PersistenceId { get; set; }
+            public long SequenceNumber { get; set; }
+            public byte[] Message { get; set; }
+            public bool Deleted { get; set; }
+            public string Manifest { get; set; }
+            public long Timestamp { get; set; }
+            public int? Identifier { get; set; }
+            public string? WriterUuid { get; set; }
+        }
```
to11mtm, line 463: Easier to use a buffer class for the AsQueryable due to mapping on `JournalRow` (namely, we kinda need to have a convert stage to avoid linq2db getting confused about what we want to insert, I _think,)_ but maybe it should be moved....

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+
+                    tagDict[(journalRow.PersistenceId, journalRow.SequenceNumber)]
+                        = journalRow.TagArray;
+                    thisInsTagSize = thisInsTagSize + journalRow.TagArray.Length;
+                    currRows++;
+                    // ByteLength *2
+                    //  + 2048 for padding
+                    // (i.e. Serializer manifests, persistence IDs, sequence numbers etc.)
+                    currBytes +=
+                        (journalRow.Message.Length * 2 + 2048);
```
to11mtm, line 623: 2048 felt like safe value for 

 - Serializer Manifest (main concern) + Identifier
 - Persistence ID (Most RDBMS will be unhappy if this is too large on a UK)
 - Sequence Number
 - Timestamp Data
 - WriterUUID
 - Deleted flag

At least averaged out? I say that cause I've seen lots of cases where SerManifest can get big (e.x. heavy generics involving deep namespaces).

1024 is probably more real, but I figured at least in this case the overpadding is not hurting us much regardless..

src/Akka.Persistence.Sql/Journal/Dao/Linq2DbHacks.cs
```diff
@@ -0,0 +1,111 @@
...
+using LinqToDB.Internal.Async;
+using LinqToDB.Internal.Linq;
+
+namespace Akka.Persistence.Sql.Journal.Dao
+{
+    public class InstantHandleAttribute : Attribute
+    {
+        
+    }
+    public static class Linq2DbHacks
```
to11mtm: This class is possibly controversial, but the way it wires up appears to be WAYYY more efficient than the normal `InsertWithOutputAsync`, possibly because it's doing at least a little less async wrapping sync.

I guess we could ditch it if we really needed to ;_;

src/Akka.Persistence.Sql/Utility/ChannelQueueWithBatch.cs
```diff
@@ -0,0 +1,198 @@
...
+                    }
+
+                    // Room in the batch — fold the element in.
+                    batch = _aggregate(batch, nextItem);
+                    remaining -= cost;
+                }
+
+                item = batch;
+                return true;
+            }
```
to11mtm, line 195: So, fun rabbit hole to send an agent down...

This is probably less efficient than a fully custom `Channel<TIn,TBatch>` could be _for this use case_. i.e. something where the reader could somehow fetch more than one item in a single `TryRead` from the write side. (Again, we are talking custom so the internals could talk to each other or do something like request a byte limit etc...)

src/Akka.Persistence.Sql/persistence.conf
```diff
@@ -60,6 +60,17 @@
...
+      # When true and tag-write-mode is TagTable, uses AsQueryable() + InsertWithOutputAsync()
+      # for tag inserts instead of individual round-trips per event.
+      # This can significantly improve write throughput for tagged events on supported databases.
+      # Current support - MSSQL, PostgreSql, Sqlite
+      use-tagtable-asqueryable-literal-insert = false
+
+      # Maximum SQL string length (in characters) for a single AsQueryable literal insert batch.
+      # When the accumulated SQL exceeds this limit, the batch is flushed to the database.
+      # Only effective when use-tagtable-asqueryable-literal-insert is true.
+      tagtable-asqueryable-insert-sql-length-limit = 5000000
```
to11mtm, line 79: I don't know if I like this number.....

src/Akka.Persistence.Sql/persistence.conf
```diff
@@ -60,6 +60,17 @@
       # But if you would prefer parameters, you can set this to true.
       prefer-parameters-on-multirow-insert = false
 
+      # When true and tag-write-mode is TagTable, uses AsQueryable() + InsertWithOutputAsync()
+      # for tag inserts instead of individual round-trips per event.
+      # This can significantly improve write throughput for tagged events on supported databases.
+      # Current support - MSSQL, PostgreSql, Sqlite
+      use-tagtable-asqueryable-literal-insert = false
```
to11mtm: Default the `AsQueryable` as disabled, leave it opt in for now.

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
```diff
@@ -43,7 +46,8 @@ public void Dispose()
             => _connection.Dispose();
 
         public AkkaDataConnection Clone()
-            => new(_providerName, (DataConnection)_connection.Clone());
+            => throw new NotImplementedException();
+        // TODO FIX: new(_providerName, (DataConnection)_connection.Clone());
```
to11mtm: do we just axe this or do the necessary wireup to re-implement?

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -32,12 +40,33 @@ public abstract class BaseByteArrayJournalDao : BaseJournalDaoWithReadMessages,
     {
         private readonly Flow<JournalRow, Util.Try<ReplayCompletion>, NotUsed> _deserializeFlowMapped;
         private readonly TagMode _tagWriteMode;
+        private readonly bool _useTagTableAsQueryable;
+        private readonly ArrayPool<JournalRow> JournalRowPool = ArrayPool<JournalRow>.Shared;
```
to11mtm: Not sure if should have `static` or instead be configurable...

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -55,21 +84,40 @@ protected BaseByteArrayJournalDao(
...
+            
+            // Channel + ChannelQueueWithBatch replaces
+            // Source.Queue + BatchWeighted. The input channel provides backpressure via
+            // BoundedChannelFullMode.Wait instead of dropping writes with DropNew.
+            _inputChannel = Channel.CreateBounded<WriteQueueEntry>(
+                new BoundedChannelOptions(JournalConfig.DaoConfig.BufferSize)
+                {
+                    FullMode = BoundedChannelFullMode.Wait,
+                    SingleReader = true,
+                    AllowSynchronousContinuations = false,
```
to11mtm, line 112: Interestingly, `true` didn't help here...

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+            public string? WriterUuid { get; set; }
+        }
+
+        protected async Task RunFastInsertNoEventParams(AkkaDataConnection connection, Seq<JournalRow> xs, BaseByteArrayJournalDaoConfig journalConfigDaoConfig, CancellationToken token)
+        {
+            var roundTripByteLimit = JournalConfig.DaoConfig.AsQueryableInsertSqlLengthLimit;
+            var rowLimit =
+                xs.Length > JournalConfig.DaoConfig.BatchSize
+                    ? xs.Length
+                    : JournalConfig.DaoConfig.BatchSize;
```
to11mtm, line 585: I know this is ugly as sin and yet it exists primarily because our `PersistAll` benchmarks are perhaps overly evil and unrealistic....

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+            var currRows = 0;
+            var currBytes = 0;
+            var thisInsTagSize = 0;
+            Dictionary<(string PersistenceId, long SequenceNumber), string[]> tagDict 
+                = new Dictionary<(string persistenceId, long sequenceNumber), string[]>();
+            JournalRow[]? insertList = null;
+            try
+            {
+                insertList = JournalRowPool.Rent(rowLimit);
+                foreach (var journalRow in xs)
```
to11mtm, line 594: This whole thing is more or less a remix on linq2db BulkCopy's `MultipleRowsHelper` and calling logic.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                                Message = input.Message,
+                                Deleted = input.Deleted,
+                                Manifest = input.Manifest,
+                                Timestamp = input.Timestamp,
+                                Identifier = input.Identifier,
+                            },
+                        (inserted) => new { inserted.Ordering, inserted.PersistenceId, inserted.SequenceNumber }, token)
+                );
+            var insertList = new List<JournalTagRow>(thisInsTagSize);
+            foreach (var ir in inserted)
```
to11mtm: Not sure if we are worried about `OrderBy` for Ordering here? Should be fine since in transaction but at least wanted to call out in case we wanted paranoia here.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
```diff
@@ -341,22 +395,197 @@ await ConnectionFactory.ExecuteWithTransactionAsync(
...
+                                SequenceNumber = input.SequenceNumber,
+                                Message = input.Message,
+                                Deleted = input.Deleted,
+                                Manifest = input.Manifest,
+                                Timestamp = input.Timestamp,
+                                Identifier = input.Identifier,
+                            },
+                        (inserted) => new { inserted.Ordering, inserted.PersistenceId, inserted.SequenceNumber }, token)
+                );
+            var insertList = new List<JournalTagRow>(thisInsTagSize);
```
to11mtm: thought about another ArrayPool here but was worried I'd get yelled at cause I already went too far.

src/Akka.Persistence.Sql/Config/BaseByteArrayJournalDaoConfig.cs
```diff
@@ -21,6 +21,8 @@ public BaseByteArrayJournalDaoConfig(Configuration.Config config)
             Parallelism = config.GetInt("parallelism", 2);
             MaxRowByRowSize = config.GetInt("max-row-by-row-size", 100);
             SqlCommonCompatibilityMode = config.GetBoolean("delete-compatibility-mode");
+            UseTagTableAsQueryableLiteralInsert = config.GetBoolean("use-tagtable-asqueryable-literal-insert", false);
+            AsQueryableInsertSqlLengthLimit = config.GetInt("tagtable-asqueryable-insert-sql-length-limit", 5_000_000);
```
to11mtm, line 24: mentioned elsewhere but I am worried this number is too big.

One option would be cribbing L2db's per-provider BulkCopy limits, however I am 90% sure that this number _is_ within range of DBs we officially support; Definitely MSSQL, PG, and SQLite (which are the only ones that support the pattern.....)

However It's possible/likely that a smaller limit is more beneficial. Alas my SSD has taken a beating even putting this together and I didn't get to test variations in depth.

to11mtm (commented):
Self review, mostly notes for others. Some of which refer to portions that will be reviewed upon acknowledgement.

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
Arkatufus: Will axing it affect runtime behavior? I have a feeling that we should be gentler with this because this is a public API. Maybe use deprecated attribute and implement our own cloning function for at least a few version?

src/Akka.Persistence.Sql/Config/BaseByteArrayJournalDaoConfig.cs
Arkatufus: One question, is this feature backward compatible with older databases? We got sniped earlier with STRING_AGG not supported in MSSQL 2016, just wanted to make sure this can still run on older database versions.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
Arkatufus: Should it ever be closed? Isn't the journal suppose to be alive until the actor system is terminated?

Arkatufus (changes requested):
Left a few questions

src/Akka.Persistence.Sql/Config/BaseByteArrayJournalDaoConfig.cs
to11mtm: ... It _should_ be, at least as far as the databases I noted having support (And coded for in the logic in the DAO accordingly when we read the config...)

For PG `INSERT..RETURNING` is supported since 8.2; for MSSQL it's since 2005 I think? 

For SQLite it's version 3.35.0, which is a bit newer than PG 8.2 or MSSQL 2005... I guess the concern there would be if folks weren't keeping their SQLite up to date (on the other hand it's probably the most trivial to update...)

Edited for Clarity:

SQLite 3.35.0 was released in 2021, Any version of `SQLitePCL.Raw` after 2.1.0 (Jun 2022) will support it.

PG 8.2 was released in 2006, but also Linq2Db only officially supports 9.2 or newer regardless.

That said I understand the concern, thus it is still 'opt-in' which at least allows people the ability to test it in an environment just in case.

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
to11mtm: Yeah, it will make the connection object a little larger but we can probably keep it to either a reference to the Factory or a `DataOptions` object to do it that way. I'll work on fixing this. and hopefully what I go with is okay.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
to11mtm: I'm probably overthinking it; I guess in my head it would be cleaner to provide a signal to close the input queue on Actor system termination?

But again, probably overthinking it.

src/Akka.Persistence.Sql/Db/AkkaDataConnection.cs
to11mtm: Actually, this gets tricky in that...

If we are considering the constructor as also part of public API, we are unfortunately cornered.

There's no reliable way to 'clone' a connection with what this class currently gets. =/

We have to add a constructor arg, one way or another to get `Clone` to work.

I guess it's up to you and @Aaronontheweb how we want to rip the bandaid off...

I will note that the _cleanest_ option long-term would be for the ctor to just take the `DataOptions` in instead... then the ctor can make the new connection and keep the `DataOptions` for cloning.

Edit to add: In my defense I don't think this class existed when this all started😅

to11mtm: Actually I'm dumb (except for the thing where, the solution is so simple I'm surprised they deprecated the clone API?)

to11mtm:
The biggest concern requiring code changes (i.e. `Clone`) I actually found a good solution for, the others I hopefully answered so I re-requested review.

Also took care of build issues that came up from merge (L2DB V6 has different namespace for how we handled `STRING_AGG`).

Will also note, The custom `ChannelReader` improves performance on Heavy concurrency even in existing scenarios, PG 400 actors in CSV mode (non-tagged) gets at least a 10% throughput boost.

Other useful numbers...


| Scenario | Batch Size | Implementation | Mean (msg/s/actor) | Mean Total (msg/s) | Median (msg/s/actor) | Median Total (msg/s) | Δ Mean Total |
|---|---|---|---:|---:|---:|---:|---:|
| Tags on TagTable, no AsQueryable | PersistGroup200 | `Source.Queue` | 5.83 | 1,166.39 | 5.78 | 1,156.09 | |
| Tags on TagTable, no AsQueryable | PersistGroup200 | `ChannelQueueWithBatch` | 10.33 | 2,066.52 | 10.41 | 2,081.77 | **+77.2%** |
| Tags on TagTable, with AsQueryable | PersistGroup200 | `Source.Queue` | 8.94 | 1,788.08 | 8.88 | 1,776.13 | |
| Tags on TagTable, with AsQueryable | PersistGroup200 | `ChannelQueueWithBatch` | 15.91 | 3,181.94 | 16.00 | 3,199.69 | **+78.0%** |
| No Tags, CSV Mode | PersistGroup200 | `Source.Queue` | 122.98 | 24,596.81 | 126.44 | 25,287.42 | |
| No Tags, CSV Mode | PersistGroup200 | `ChannelQueueWithBatch` | 121.54 | 24,307.63 | 124.40 | 24,880.36 | **-1.2%** |
| No Tags, CSV Mode | PersistGroup400 | `Source.Queue` | 67.41 | 26,962.66 | 69.13 | 27,650.78 | |
| No Tags, CSV Mode | PersistGroup400 | `ChannelQueueWithBatch` | 79.36 | 31,743.98 | 80.45 | 32,179.37 | **+17.7%** |

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
Arkatufus: I think having the stream closed unexpectedly while the journal is still running is the more detrimental and dangerous case. This is a pre-existing problem though, we'd need to figure this out in the future.

src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs
Arkatufus: Might want to clean the uwu and emoji from the documentation here

src/Akka.Persistence.Sql/Journal/Dao/Linq2DbHacks.cs
Arkatufus: Maybe rename the class name to a less controversial `Linq2DbExtensions` name?

src/Akka.Persistence.Sql/Utility/ChannelQueueWithBatch.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <param name="inputReader">
+        /// The reader side of an input channel owned by the caller. Typically created
+        /// from a <c>BoundedChannel&lt;TInput&gt;</c> with <c>FullMode = Wait</c>.
+        /// The caller retains the <c>ChannelWriter&lt;TInput&gt;</c> for producing items.
+        /// </param>
+        /// <param name="maxWeight">
+        /// Maximum total cost budget for a single batch. Once accumulated cost reaches
+        /// this limit, the batch is emitted and the overflow item is parked for the
+        /// next read. Maps to the <c>batch-size</c> config value in
+        /// <c>BaseByteArrayJournalDaoConfig</c>. 🌸
```
Arkatufus: Need to remove the emoji here too

src/Akka.Persistence.Sql/persistence.conf
Arkatufus: Really not sure about this either, maybe @Aaronontheweb can chip in on this

Arkatufus (changes requested):

Arkatufus:
> Are able to use .AsQueryable() + .InsertWithOutputAsync(), to mitigate the performance issues around tagged event results in 'one event roundtrip, bulkinsert tags for that event', this will for supporting databases:
>
> 1. Build a lookup for post event insert for what to use on tag insert
> 2. Append entries to a collection and increment counters
> 3. When counters go above threshold, flush the entries and the lookups based on joining from .InsertWithOutputAsync.
>
> Basically, it's a simplified version of how Linq2Db's MultipleRows BulkCopy works, we just have to do more to manually define the round trips.

Just to make sure I understand the concept. The bulk insert is still done atomically per persist operation and not defered until the counter threshold elapsed, correct?

to11mtm:
> > Are able to use .AsQueryable() + .InsertWithOutputAsync(), to mitigate the performance issues around tagged event results in 'one event roundtrip, bulkinsert tags for that event', this will for supporting databases:
> > 
> > 1. Build a lookup for post event insert for what to use on tag insert
> > 2. Append entries to a collection and increment counters
> > 3. When counters go above threshold, flush the entries and the lookups based on joining from .InsertWithOutputAsync.
> > 
> > Basically, it's a simplified version of how Linq2Db's MultipleRows BulkCopy works, we just have to do more to manually define the round trips.
> 
> Just to make sure I understand the concept. The bulk insert is still done atomically per persist operation and not defered until the counter threshold elapsed, correct?

Not sure entirely what you're asking as far as 'until the counter threshold elapsed' here...

I'll try to answer though; this is still _eager_ as far as writing (i.e. if we are under counter threshold we still insert what we've got ASAP), it's just that instead of us doing `Insert JournalRow returning Inserted ordering->Insert TagRows for JournalRow` multiple times in a single transaction, we are doing a smaller number of `Insert JournalRows getting back orderings in a correlated fashion->Insert Tagrows for that correlated set via bulkcopy` in a single transaction.

Arkatufus:
Ah, got it, thanks.

